PR #1284, opened 2023-11-02
fix: permission issue on create actions and improved permission performance

davenewza:
## Permission issue on create

Because of how we had implemented permission checking, the following action was NOT authorising correctly.  This has been fixed.

```gql
create createProduct() with (name, barcode, productCode, brand.id) {
    @permission(expression: ctx.identity.user in product.brand.editors)
}
```

## Performance improvements on permission

For larger datasets, we will see a significant performance improvement in the way we check permissions.  Previously we would have seen a major degradation (even to unusable levels).

### DIFF
--- a/runtime/actions/authorization.go
+++ b/runtime/actions/authorization.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sort"
 	"strings"
 
 	"github.com/samber/lo"
@@ -267,20 +266,4 @@ func getEmailAndDomain(ctx context.Context) (email string, domain string, verifi
 	segments := strings.Split(identity.Email, "@")
 	domain = segments[1]
 	return identity.Email, domain, identity.EmailVerified, nil
-}
-
-func compare(a, b []string) bool {
-	if len(a) != len(b) {
-		return false
-	}
-
-	sort.Strings(a)
-	sort.Strings(b)
-
-	for i, v := range a {
-		if v != b[i] {
-			return false
-		}
-	}
-	return true
 }
--- a/runtime/actions/authorization.go
+++ b/runtime/actions/authorization.go
@@ -3,6 +3,7 @@ package actions
 import (
 	"context"
 	"errors"
+	"fmt"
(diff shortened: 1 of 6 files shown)


runtime/actions/query.go
```diff
@@ -1147,7 +1148,18 @@ func (query *QueryBuilder) generateConditionTemplate(lhs *QueryOperand, operator
...
 			inPlaceholders := []string{}
-			inValues := rhs.value.([]any)
+
+			// Check that rhs is a slice
+			slice := reflect.ValueOf(rhs.value)
+			if slice.Kind() != reflect.Slice || slice.IsNil() {
+				return "", nil, errors.New("rhs operand in sql condition must be a slice")
+			}
+
+			// Safely map rhs slice to []any
```
louisaspicer, line 1158: Out of interest, why do you need to map to []any?

runtime/actions/authorization.go
```diff
@@ -263,19 +267,3 @@ func getEmailAndDomain(ctx context.Context) (email string, domain string, verifi
 	domain = segments[1]
 	return identity.Email, domain, identity.EmailVerified, nil
 }
-
-func compare(a, b []string) bool {
```
louisaspicer, line 267: Is this what caused bad performance?

louisaspicer (approved):

runtime/actions/authorization.go
davenewza: Nope.  The reason for the bad performance was because our permissions DB query was returning all of the IDs of a which the user has access too (regardless of the query being made).  So, if there were 1 million rows which the user was allowed to see, then 1 million IDs would be returned for every permission check

runtime/actions/query.go
davenewza: I cant remember right now 🤔 .  I'll come back to this and write a nice comment 👍

runtime/actions/authorization.go
davenewza: And now we only return a count.

PR #1285, opened 2023-11-02
fix: display full output path when generating a client 

tomfrew:
If generated a client with an output dir flag e.g. `-o ../my/frontend` we didn't show the correct output path. 

New output

**Single file client**
```
📦 Generating client SDK..                         
✅ All done!                                       
                                                   
The following files were generated:                
                                                   
../customDir/@teamkeel/client/keelClient.ts  
```

**Package client**
```
📦 Generating client SDK..                         
✅ All done!                                       
                                                   
The following files were generated:                
                                                   
../customDir/core.ts   
../customDir/@teamkeel/client/index.ts  
../customDir/@teamkeel/client/types.ts  
../customDir/@teamkeel/client/package.ts
```

### DIFF
--- a/cmd/program/client.go
+++ b/cmd/program/client.go
@@ -206,9 +206,17 @@ func (m *GenerateClientModel) renderGenerate() string {
 				parts := strings.Split(generatedFile.Path, "/")
 				prePath := filepath.Join(parts[0 : len(parts)-1]...)
 
+				if m.OutputDir != "" {
+					if prePath != "" {
+						prePath = fmt.Sprintf("%s/%s", m.OutputDir, prePath)
+					} else {
+						prePath = m.OutputDir
+					}
+				}
+
 				b.WriteString(
 					colors.Gray(
-						fmt.Sprintf("- %s/%s%s", prePath, colors.Cyan(functionName).String(), colors.Gray(".ts").String()),
+						fmt.Sprintf("%s/%s%s", prePath, colors.Cyan(functionName).String(), colors.Gray(".ts").String()),
 					).Highlight().String(),
 				)
 				b.WriteString("\n")


adaam2:
Personally I think the client should exist in the same location as the other packages in the .build directory, using the same tsconfig paths magic, allowing you to import it as a node module in the same way as you can the others. Wdyt?

👋 by the way

tomfrew:
Hey @adaam2! 👋 

I'm not sure if the build directory is ideal for this as it's quite internal for the Keel side of things. Even if you have a mono repo style setup with co-located frontend and backend (which hopefully we'll have better support for soon) you really want the client to be part of your frontend. The .build/ directory in gitignored by default so will be missing unless you bring in a keel generate step into you frontend build pipeline.

We do want to have a smoother process for installing a generated client thought. The current setup isn't great.

adaam2:
Yeah you make a good point; it's kinda tricky to follow the same approach Prisma client uses (postinstall hook with prisma generate) if the frontend is in a different repo entirely lacking the keel schema for context

cmd/program/client.go
```diff
@@ -206,9 +206,17 @@ func (m *GenerateClientModel) renderGenerate() string {
 				parts := strings.Split(generatedFile.Path, "/")
 				prePath := filepath.Join(parts[0 : len(parts)-1]...)
 
+				if m.OutputDir != "" {
```
jonbretman, line 209: Minor point but this is a little convoluted. You can use more `filepath` methods avoid the nested if-statement.

```go
// extension (so as not to hard-code .ts)
ext := filepath.Ext(generatedFile.Path)

// filename without extension
functionName := strings.TrimSuffix(filepath.Base(generatedFile.Path), ext)
	
// directory without filename
dir := filepath.Dir(generatedFile.Path)

// destination directory
dest := filepath.Join(m.OutputDir, dir)

// Format it all
fmt.Sprintf("- %s/%s%s", dest, colors.Cyan(functionName).String(), colors.Gray(ext).String())
```

jonbretman (approved):

cmd/program/client.go
davenewza: @tomfrew @jonbretman going to merge this PR as it is approved and just hanging around

PR #1286, opened 2023-11-03
feat: support human readable schedule expressions and better validation

jonbretman:
### Summary

* Human-readable schedule expressions like **"every 10 minutes"** or **"every monday at 9am"** with strict validation
* Validation for cron-syntax with conversion from unix-style to what EventBridge expects
* Validation that there is only one `@schedule` attribute in a job - we only support one

### Human-readable schedules

The following formats are supported:

* `every N minutes`
* `every N minutes from T to T`
* `every N hours`
* `every N hours from T to T`
* `every D at T`

Where `N` is an integer, `T` is a 12-hour time, and `D` is a day. 

In the `every D at T` format both `D` and `T` can be lists of values, for example **"every monday and friday at 9am, 12pm, and 5pm"** produces the cron expression `0 9,12,17 ? * MON,FRI *`. `D` can also be `day` for everyday or `weekday` for MON-FRI.

### Cron schedules

Cron syntax is still supported but it is now strictly validated against unix-style rules with the following two exceptions:

* You can't have a value in both day-of-month and day-of-week - this is an EventBridge limitation
* You can't use a step value (e.g. `*/2`) in day-of-week - EventBridge does not support this

### Validation

Validation errors will point to the relevant part of the string where the error occurs (where possible). For example:

Screenshot 2023-11-03 at 13 09 54

### DIFF
--- a/cron/cron.go
+++ b/cron/cron.go
@@ -1,0 +1,611 @@
+package cron
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/samber/lo"
+)
+
+type Error struct {
+	Message string
+	Token   *Token
+}
+
+func (e Error) Error() string {
+	position := ""
+	if e.Token != nil {
+		position = fmt.Sprintf("(%d:%d) ", e.Token.Start, e.Token.End)
+	}
+	return fmt.Sprintf("%s%s", position, e.Message)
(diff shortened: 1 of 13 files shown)


davenewza (approved):
This is really great.  Next step is to add into the docs; I'll make a note of this

PR #1288, opened 2023-11-08
feat: token endpoint and id token authentication

davenewza:
We are to replace the `authenticate()` action and password-based authentication with a set of OAuth endpoints decoupled from the customer's APIs and Keel schema. This means much better support to integrate with 3rd party OAuth providers (such as Google and Facebook), alignment with industry guidelines and standards, and a more secure environment for customer's data and applications.  

This PR kicks off some of this work with the following...

Token Endpoint
===
The token endpoint, located at `/auth/token`, will now handle authentication requests.  It currently accepts the `token-exchange` grant for ID Token authentication (see next section).  The token endpoint will always return the following JSON response:

```
{
    "access_token": "eyJhbGciOiJSUzI...",
    "token_type": "Bearer",
    "expires_in": 86400,
    "refresh_token": "FejvZRD2jBLdoQhvpj"
}
```
**Note:** refresh tokens are not available yet.

ID Token Authentication
===
Users can authenticate with a valid ID token from a registered OpenID Connect provider without needing to configure a client ID or secret with Keel.

This performs all the necessary steps to validate and verify the legitimacy of the ID token.

Access tokens issued from the token endpoint are now decoupled from the original issuer, and they should be used in tandem with refresh tokens.

Testability
===
The majority of my effort here was making this work testable.  I've written a OIDC test server, built on top of @tomfrew's previous work.  This lets us mock a provider for testing purposes.  I think this will become very important as we start to support more flows.  Auth cannot let our users down.

Telemetry
===

The telemetry was quite noisy before.  It's a bit lighter now:
image

Outstanding
===
 - `aud` verification during ID Token authentication
 - Refresh tokens and refresh token rotation
 - keelconfig.yaml configuration for OIDC providers
 - Add the standard OIDC user claims to `Identity` and update them on authentication
 - Fetching data from the OIDC userinfo endpoint
 - The `createIfNotExists` concept in keelconfig.yaml
 - Caching the OpenID Connect profiles and JWKS public set

Questions to address
===

 - Should it be a minimum requirement that ID tokens provide us with an `email` claim?  
 - Should the issuer of the access tokens be the environment's domain?  Or just `https://keel.so`?


runtime/apis/authapi/token_endpoint.go
```diff
@@ -0,0 +1,214 @@
...
+
+			if idTokenRaw == "" {
+				return common.NewJsonResponse(http.StatusBadRequest, &TokenErrorResponse{
+					Error:            TokenEndpointInvalidRequest,
+					ErrorDescription: "the ID token in the subject_token field cannot be an empty string",
+				}, nil)
+			}
+
+			span.SetAttributes(
+				attribute.String(ArgSubjectToken, idTokenRaw),
```
jonbretman, line 128: we shouldn't put sensitive values in traces

davenewza: Oh, good catch!  I wasn't thinking

runtime/oauth/authentication.go
```diff
@@ -0,0 +1,66 @@
...
+
+	// Establishes new OIDC provider. This will call the providers discovery endpoint.
+	provider, err := oidc.NewProvider(ctx, issuer)
+	if err != nil {
+		return nil, err
+	}
+	span.AddEvent("Provider's ODIC config fetched")
+
+	// TODO: what are we missing by skipping the client ID check?
+	verifier := provider.Verifier(&oidc.Config{SkipClientIDCheck: true})
```
jonbretman, line 55: Might be more helpful to comment _why_ you set this to true in the first place?

runtime/runtime.go
```diff
@@ -37,40 +38,40 @@ func GetVersion() string {
 }
 
 func NewHttpHandler(currSchema *proto.Schema) http.Handler {
-	var handler common.ApiHandlerFunc
+	var apiHandler common.ApiHandlerFunc
```
jonbretman, line 41: might want to rename this interface to just `HandlerFunc`

runtime/oauth/authentication.go
davenewza: Done.  So this will be enabled once we're capturing the customer's client ID

PR #1289, opened 2023-11-09
fix: validation for unsupported job input types

davenewza:
Schema validation for unsupported job input types
===
Job inputs must be one of the built-in Keel types or a user-defined enum, which is now being validated for.  

NOTE:  What is unclear to me right now is how well we support arrays.  I believe we do, but it needs some investigating, testing and documenting.  I've made [a ticket for this](https://linear.app/keel/issue/KE-1252/job-array-inputs).  I've allowed array inputs for now.

image


schema/testdata/validation_fields_invalid_types/schema.keel
davenewza, line 1: Model field inputs was scantily tested, so just bolstering it up a bit here

davenewza:
I am going to merge this without approval.  Would like to get this included in the release, especially since we'll be highlighting jobs in the release notes

PR #1291, opened 2023-11-09
fix: labelled inputs sometimes validated against for functions

davenewza:
Fixed issue experienced by Jay, whereby labelled inputs were not being permitted on function actions.


schema/validation/unused_inputs.go
```diff
@@ -44,10 +38,10 @@ func UnusedInputRule(_ []*parser.AST, errs *errorhandling.ValidationErrors) Visi
 				)
 			}
 
-			unused = map[string]*parser.NameNode{}
+			isAction = false
```
jonbretman, line 41: This should be on the first line of this hook - we always want to set this to false when we leave

davenewza: Whoops.  Im running on fumes today!

jonbretman (approved):

PR #1292, opened 2023-11-09
fix: set unique to true for primary key fields in the proto.schema

louisaspicer:
Fixes the proto schema to have primary key fields set as unique. It is misleading if the flag is set to false.


migrations/migrations.go
```diff
@@ -289,7 +289,7 @@ func New(ctx context.Context, schema *proto.Schema, database db.Database) (*Migr
 				return c.TableName == tableName && c.ConstraintType == "u" && len(c.ConstrainedColumns) == 1 && c.ConstrainedColumns[0] == int64(column.ColumnNum)
 			})
 
-			if field.Unique && !hasUniqueConstraint {
+			if field.Unique && !field.PrimaryKey && !hasUniqueConstraint {
```
davenewza, line 292: 👍

davenewza (approved):
Looks good to me

PR #1293, opened 2023-11-09
fix: return record or null from findOne method in QueryBuilder

jonbretman:
I was actually trying to fix a different bug but hit this on the way.

Crazy thing was we had a test asserting that it was broken... 🙃

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,8 +1,3 @@
 module.exports = {
-  extends: ['@commitlint/config-conventional'],
-  rules: {
-    'body-max-length': [2, 'always', 300],
-    'body-case': [1, "always", "sentence-case"],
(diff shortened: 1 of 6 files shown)


commitlint.config.js
```diff
@@ -1,8 +1,3 @@
 module.exports = {
-  extends: ['@commitlint/config-conventional'],
-  rules: {
-    'body-max-length': [2, 'always', 300],
-    'body-case': [1, "always", "sentence-case"],
-    'subject-case': [2, "always", ["sentence-case", "lower-case"]]
```
jonbretman, line 6: I refuse to not be allowed to use capital letters in my commits 😄

louisaspicer: 😂 I just hit this. Trying to say proto.Schema in my commit

davenewza (approved):

PR #1294, opened 2023-11-10
feat: refresh tokens & refresh grant

davenewza:
# Refresh Tokens

We now issue a new refresh token from the token endpoint with every successful authentication.  

```json
{
  "access_token": "eyJhbGciO...",
  "token_type": "bearer",
  "expires_in": 86400,
  "refresh_token": "WQl1q2yAPzd5SrLnpZSKarFLPJFnnLRIfViJG0BfL7A76VNUnqXgkHiylpZMYeBM"
}
```

The refresh token can be used to mint a new access token without getting the user involved in reauthenticating.  Refresh tokens also give us the means to have stateless access token validation but still giving us the option to revoke access to a user if we need (the recommendation here is actually to have a short-lived access token, which we can spell out in the docs).

The refresh token is opaque (and not a structured JWT) and has a server-side configured expiry.  A refresh token is revoked by deleting it from storage.

### Storage and performance

We use symmetrical SHA3 hashing when storing the tokens in the database.  This is without a salt, which means we can easily compare hash values when looking up tokens.

Querying this table is going to be suboptimal unfortunately - it's a trade-off we need to make if we don't want to store indexed plaintext tokens.  There are ways to make this more efficient, but I think we'll be fine for now.

## Refresh Grant

The token endpoint now supports the `refresh_token` grant, which will issue a new access token.  The response from the token endpoint will be exactly the same as with any other grant.

## Refresh Token Rotation

Basically, this means that when refreshing the access token, we also issue a new refresh token alongside it, and we revoke the original refresh token.  

> Refresh tokens in SPAs was considered insecure because they typically have long lifetimes and browser storage is susceptible to token theft. However, there are acceptable mitigations outlined in the most recent [OAuth2 BCP for browser-based applications](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-browser-based-apps-05) stipulating that refresh tokens may be used provided that they adhere to specific recommendations, including having a maximum lifetime and requiring rotation to protect against malicious re-use.

Enabled by default, but we may want to make this configurable for clients which don't want a rotating refresh token.

## Things to do

 - Configuration for access token and refresh token expiry time (will also let me write some more tests)
 - Configuration to enable/disable refresh token rotation
 - Revoke endpoint
 - Find a way to clean out expired tokens from database

## Telemetry Screenshots

**Token Exchange:**

image

**Token Refresh:**

image


runtime/oauth/id_token.go
```diff
@@ -31,3 +42,61 @@ type UserClaims struct {
 	PhoneNumber         string `json:"phone_number,omitempty"`
 	PhoneNumberVerified bool   `json:"phone_number_verified,omitempty"`
 }
+
+var tracer = otel.Tracer("github.com/teamkeel/keel/runtime")
+
+// VerifyIdToken will verify the ID token from an OpenID Connect provider.
```
davenewza, line 48: Nothing changed here - just moved the function

runtime/oauth/refresh_token.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	now := time.Now().UTC()
+	expiry := now.Add(defaultRefreshTokenExpiry)
+
+	sql := `
+		INSERT INTO 
+			keel_refresh_token (token, identity_id, expires_at, created_at) 
+		VALUES 
+			(crypt(?, gen_salt('md5')), ?, ?, ?)`
```
jonbretman, line 38: Why not use the same approach we use for passwords and encrypt in Go? https://github.com/teamkeel/keel/blob/d21ba7dbf9cc6f371d72d65c56d0948993e8eb04/runtime/actions/authenticate.go#L113-L113

davenewza: Bcrypt uses a salt when hashing which means that two hashes of the same token will not be the same.  This means we can't simply compare the hash varchars in a database query.  We would need to use `bcrypt.CompareHashAndPassword` on each token from the database until we find our match (and, at worst, having to read the entire table into memory), because `bcrypt.CompareHashAndPassword` will use the salt when comparing the hash with the new token.

BUT, your question got me researching.   Im no security expert, but I think it's completely unnecessary to hash with salt for tokens (which are cryptographically generated), which means we can use a single round of SHA3 hashing which _always_ produces the same hash.  This means we can compare the hash values, which will be tremendously more performant (as we can now index them!)

davenewza: Im very glad you asked, because this entirely solves the performance challenge

PR #1296, opened 2023-11-13
feat: keelconfig auth configuration

davenewza:
## Auth configuration in keelconfig.yaml

The following can be configured:
 - Access and refresh token expiries
 - Built-in providers (`google` supported in this PR)
 - Custom `oidc` providers for ID token and OAuth authentication
 - Custom `oauth` providers for OAuth authentication only

Each type has it's own set of requirements, but all require a `name` and a `clientId`

```yaml
auth:
  tokens:
    # In seconds
    accessTokenExpiry: 3600
    refreshTokenExpiry: 604800
    
  providers:
    # Built-in Google provider
    - type: google
      name: GoogleClient1
      clientId: jjshflsis2ss23

    # Built-in Google provider
    - type: google
      name: GoogleAdmin
      clientId: 23f223twefgwfw

    # Custom OIDC
    - type: oidc
      name: Auth0
      issuerUrl: 'https://dev-skhlutl45lbqkvhv.us.auth0.com'
      clientId: 'kasj28fnq09ak'

    # Custom OAuth
    - type: oauth
      name: Github
      clientId: hfjuw983h1hfsdf
      authorizationUrl: https://github.com/auth
      tokenUrl: https://github.com/token
```

### Telemetry

If a issuer is not registered, then the client will get the standard 401:

```json
{
    "error": "invalid_client",
    "error_description": "possible causes may be that the id token is invalid, has expired, or has insufficient claims"
}
```

But the tracing will capture the underlying reason properly:

image

### Outstanding

 - Per environment/slug configuration
 - Disable refresh token rotation by config item
 - How we do the client secret thing for OAuth providers

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -254,7 +254,7 @@ func Validate(config *ProjectConfig) *ConfigErrors {
 	}
 
 	missingProviderNames := findAuthProviderMissingName(config.Auth.Providers)
-	for i, _ := range missingProviderNames {
+	for i := range missingProviderNames {
 		errors = append(errors, &ConfigError{
 			Type:    "missing",
 			Message: fmt.Sprintf(ConfigAuthProviderMissingFieldAtIndexErrorString, i, "name"),
--- a/config/config.go
+++ b/config/config.go
@@ -239,14 +239,14 @@ func Validate(config *ProjectConfig) *ConfigErrors {
 		}
 	}
 
-	if config.Auth.Tokens.AccessTokenExpiry <= 0 {
+	if config.Auth.Tokens != nil && config.Auth.Tokens.AccessTokenExpiry <= 0 {
 		errors = append(errors, &ConfigError{
 			Type:    "invalid",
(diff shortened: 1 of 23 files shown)


config/config.go
```diff
@@ -232,6 +239,94 @@ func Validate(config *ProjectConfig) *ConfigErrors {
...
+	}
+
+	if config.Auth.Tokens != nil && config.Auth.Tokens.RefreshTokenExpiry <= 0 {
+		errors = append(errors, &ConfigError{
+			Type:    "invalid",
+			Message: fmt.Sprintf(ConfigAuthTokenExpiryMustBePositive, "refresh", "refreshTokenExpiry"),
+		})
+	}
+
+	missingProviderNames := findAuthProviderMissingName(config.Auth.Providers)
```
jonbretman, line 256: I feel like we could do all this validation using a JSON schema. Not for this PR but something to consider.

davenewza: Totally agree.  I should have done that from the start

jonbretman (approved):
LGTM

PR #1297, opened 2023-11-13
fix: id token verification performs client id check

davenewza:
### Client ID verification for ID token verification

An ID token will always have the OIDC server's client ID in the `sub` claim, and it is important that the ID token verification steps include checking the client ID.  Otherwise it's possible that the ID token was obtained from another application and wasn't intended for this app.

### Supports providers with the same issuer URL

It's possible for customers to configure multiple clients from the same issuer, so this PR also supports that now.

### DIFF
--- a/runtime/oauth/id_token.go
+++ b/runtime/oauth/id_token.go
@@ -66,7 +66,12 @@ func VerifyIdToken(ctx context.Context, idTokenRaw string) (*oidc.IDToken, error
 		return nil, err
 	}
 
-	if !authConfig.HasOidcIssuer(issuer) {
+	hasIssuer, err := authConfig.HasOidcIssuer(issuer)
+	if err != nil {
+		return nil, err
+	}
+
+	if !hasIssuer {
 		return nil, fmt.Errorf("issuer %s not registered to authenticate on this server", issuer)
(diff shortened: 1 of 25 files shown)


runtime/oauth/id_token.go
```diff
@@ -82,17 +82,25 @@ func VerifyIdToken(ctx context.Context, idTokenRaw string) (*oidc.IDToken, error
...
+			ClientID: p.ClientId,
+		}
+
+		// Verify that the ID token legitimately was signed by the provider and that it has not expired
+		idToken, err := provider.Verifier(oidcConfig).Verify(ctx, idTokenRaw)
+		if err != nil {
+			verificationErrs = errors.Join(err, verificationErrs)
+		} else {
+			span.AddEvent("ID Token verified")
+			return idToken, nil
```
louisaspicer, line 101: Does this need to verify all the providers?

davenewza: It's only testing against the providers which have the same issuer as the ID token:
`providers, err := authConfig.GetOidcProvidersByIssuer(issuer)`

louisaspicer (approved):

PR #1298, opened 2023-11-14
feat: refresh token rotation configuration & github and facebook providers

davenewza:
### `facebook` and `gitlab` as built-in OpenID Connect providers

No need to provide issuer, token or auth URLs for Facebook and GitLab.  They only need to be specified like this in the config:

```yaml
  providers:
    - type: facebook
      name: fb_myapp
      clientId: 1234567
```

### Refresh Token Rotation configureable

For some clients it may be important to disable refresh token rotation.  When rotation is disabled, the original refresh token will be returned when minting a new access token.  This can be done in the config as follows:

```yaml
  tokens:
    accessTokenExpiry: 3600
    refreshTokenExpiry: 86400
    refreshTokenRotationEnabled: false. # <-- disable rotation
```

### Other changes

 - Improved reading from the config and dealing with default values
 - Cleaned up some existing tests which made no more sense

### DIFF
--- a/config/auth.go
+++ b/config/auth.go
@@ -41,9 +41,9 @@ type AuthConfig struct {
 }
 
 type TokensConfig struct {
-	AccessTokenExpiry    *int  `yaml:"accessTokenExpiry,omitempty"`
-	RefreshTokenExpiry   *int  `yaml:"refreshTokenExpiry,omitempty"`
-	RefreshTokenRotation *bool `yaml:"refreshTokenRotation,omitempty"`
+	AccessTokenExpiry           *int  `yaml:"accessTokenExpiry,omitempty"`
+	RefreshTokenExpiry          *int  `yaml:"refreshTokenExpiry,omitempty"`
+	RefreshTokenRotationEnabled *bool `yaml:"refreshTokenRotationEnabled,omitempty"`
 }
 
 type Provider struct {
@@ -75,8 +75,8 @@ func (c *AuthConfig) RefreshTokenExpiryOrDefault() time.Duration {
 
 // RefreshTokenRotationEnabled retrieves the configured or default refresh token rotation
 func (c *AuthConfig) RefreshTokenRotationEnabled() bool {
-	if c.Tokens.RefreshTokenRotation != nil {
-		return *c.Tokens.RefreshTokenRotation
+	if c.Tokens.RefreshTokenRotationEnabled != nil {
+		return *c.Tokens.RefreshTokenRotationEnabled
 	} else {
 		return true
 	}
--- a/config/auth.go
+++ b/config/auth.go
@@ -22,9 +22,8 @@ var (
 )
 
 type AuthConfig struct {
-	Tokens              *TokensConfig `yaml:"tokens"`
-	AllowAnyOidcIssuers bool          `yaml:"allowAnyOidcIssuer"`
-	Providers           []Provider    `yaml:"providers"`
+	Tokens    *TokensConfig `yaml:"tokens"`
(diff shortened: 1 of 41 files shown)


runtime/apis/authapi/token_endpoint.go
```diff
@@ -66,6 +67,15 @@ func TokenEndpointHandler(schema *proto.Schema) common.HandlerFunc {
 		ctx, span := tracer.Start(r.Context(), "Token Endpoint")
 		defer span.End()
 
+		var identityId string
+		var refreshToken string
+
+		config, err := runtimectx.GetOAuthConfig(ctx)
+		if err != nil {
+			span.RecordError(err)
```
jonbretman, line 75: with stack trace?

runtime/apis/authapi/token_endpoint.go
```diff
@@ -111,10 +121,24 @@ func TokenEndpointHandler(schema *proto.Schema) common.HandlerFunc {
...
-			isValid, newRefreshToken, identityId, err := oauth.RotateRefreshToken(ctx, refreshTokenRaw)
-			if err != nil {
-				span.RecordError(err)
-				return common.NewJsonResponse(http.StatusInternalServerError, nil, nil)
+			var isValid bool
+			if config.RefreshTokenRotationEnabled() {
+				// Rotate and revoke this refresh token, and mint a new one.
+				isValid, refreshToken, identityId, err = oauth.RotateRefreshToken(ctx, refreshTokenRaw)
+				if err != nil {
+					span.RecordError(err)
```
jonbretman, line 129: with stack trace?

runtime/apis/authapi/token_endpoint.go
```diff
@@ -217,35 +226,37 @@ func TokenEndpointHandler(schema *proto.Schema) common.HandlerFunc {
...
 			return common.NewJsonResponse(http.StatusBadRequest, &TokenErrorResponse{
 				Error:            TokenEndpointUnsupportedGrantType,
 				ErrorDescription: "the only supported grants are 'refresh_token' and 'token_exchange'",
 			}, nil)
 		}
+
+		// Generate a new access token for this identity.
+		accessTokenRaw, expiresIn, err := oauth.GenerateAccessToken(ctx, identityId)
+		if err != nil {
+			span.RecordError(err)
```
jonbretman, line 248: why do you hate stack traces? 😄

config/auth.go
```diff
@@ -40,6 +54,33 @@ type Provider struct {
 	AuthorizationUrl string `yaml:"authorizationUrl"`
 }
 
+// AccessTokenExpiryOrDefault retrieves the configured or default access token expiry
+func (c *AuthConfig) AccessTokenExpiryOrDefault() time.Duration {
```
jonbretman, line 58: Kind of feel like the `OrDefault` suffix on these methods is not needed. It's explained in the doc comment, and typed to always return a `time.Duration`.

runtime/apis/authapi/token_endpoint.go
davenewza: Oops yes.  I've also set `span.SetStatus(codes.Error, err.Error())` for these (only the "internal server" errors though)

runtime/apis/authapi/token_endpoint.go
davenewza: I swear I really don't 😭

config/auth.go
davenewza: Sounds good - I've fixed them

PR #1299, opened 2023-11-14
chore: refactor function tests and fix bugs

jonbretman:
### Bug fix: non-model inputs
The main bug that is fixed here which affected all action types was that if your function had any non-model inputs your function would error if the default query was run. For example if you had a schema like this:

```
model Thing {
  actions {
    get getThing(id, something: Text) @function
  }
}
```

The database query that would get generated would try to do `where something = ?` even though `something` is not a model field.

**This is now fixed across all action types**.

### Improvement: types
A lot of the hook types were funny about whether the hook could return a `T` or a `Promise<T>` - **all hooks can now be sync or async.**

### Hooks order / arguments

For `update` functions these are now the hook types for a model called `Book`:
```ts
export type UpdateBookBeforeQueryHooks = {
  beforeQuery?: (
    ctx: ContextAPI,
    inputs: UpdateBookBeforeQueryInput,
    query: BookQueryBuilder
  ) => Promise<Book | BookQueryBuilder> | Book | BookQueryBuilder;

  beforeWrite?: (
    ctx: ContextAPI,
    inputs: UpdateBookBeforeQueryInput,
    values: UpdateBookBeforeQueryValues,
    record: Book
  ) => Promise<Partial<Book>> | Partial<Book>;

  afterWrite?: (
    ctx: ContextAPI,
    inputs: UpdateBookBeforeQueryInput,
    data: Book
  ) => Promise<Book | void> | Book | void;
}
```

The changes here are:
* `beforeQuery` can return a query builder
* There is no `afterQuery` hook
* `beforeWrite` is now passed the current record as a 4th argument, so it can be used to do updates based on existing values
* `afterWrite` can return a record if you want to change what is returned from the API without changing what is in the database. Basically this is the write version of `afterQuery`.

### Tests
I've renamed some of the tests to just be `functions_something` rather than `with_custom_functions_something`. Also consolidated (basically rewrote) the `with_custom_functions` and `with_custom_functions_hooks` tests - these are now just `functions_hooks`.


davenewza:
@jonbretman I'm hoping to do a release tomorrow morning, but may need your help as I am a little uncertain of what to say about breaking changes.  From what I can tell, it's just that `afterQuery` has been removed?

davenewza:
Maybe we should start a new branch on the docs for these changes and the release notes

node/codegen_test.go
```diff
@@ -1578,351 +1578,6 @@ export declare function SendWelcomeEmail(fn: (ctx: SubscriberContextAPI, event:
 	})
 }
 
-func TestWriteFunctionWrapperType(t *testing.T) {
```
davenewza, line 1581: How do you feel about new codegen tests?

davenewza: Im sure you are pretty tested out by now 😝

jonbretman: I intentionally removed these as I don't think they add much value. The integration tests make sure the code (and types) work correctly.

jonbretman:
> @jonbretman I'm hoping to do a release tomorrow morning, but may need your help as I am a little uncertain of what to say about breaking changes. From what I can tell, it's just that `afterQuery` has been removed?

Yep let's chat about docs / release notes.

PR #1303, opened 2023-11-17
chore: refactor some of the sdk codegen for functions

jonbretman:
I see this is as a stepping stone to moving most of the function hooks stuff into the functions-runtime package, but I think until the API is 100% stable it's better in codegen as that avoids any issues with version drift between the codegen and the npm package.


davenewza (approved):
Super

PR #1304, opened 2023-11-17
feat: support setting http response status in functions

jonbretman:
This makes it possible to set the HTTP response status from a function. This only has an effect on the HTTP JSON API, for GraphQL and JSON RPC the response status will always be `200`.

#### Simple example
```ts
import { MyFunction, permissions } from "@teamkeel/sdk";

export default MyFunction(async (ctx) => {
  permissions.allow();

  const { response } = ctx;
  response.headers.set("Location", "https://some.url");
  response.status = 301;

  return null;
});
```

#### Request Headers
In this change I've also dropped the `RequestHeaders` class as I don't think it's necessary. Instead I've updated the types to be a `Headers` object with the mutating methods dropped.

### DIFF
--- a/packages/functions-runtime/src/index.d.ts
+++ b/packages/functions-runtime/src/index.d.ts
@@ -59,6 +59,7 @@ export type ContextAPI = {
 
 export type Response = {
   headers: Headers;
+  status?: number;
 };
 
 export type PageInfo = {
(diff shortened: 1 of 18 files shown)


TejasQ:
Great! Let’s add this to the docs.

jonbretman:
> Great! Let’s add this to the docs. 
> 
> 
> 
> 

Yep I'll get a docs PR together to go with this.

packages/functions-runtime/src/index.d.ts
```diff
@@ -69,11 +70,8 @@ export type PageInfo = {
   count: number;
 };
 
-// Request headers query API
-export type RequestHeaders = {
-  get(name: string): string;
-  has(name: string): boolean;
-};
+// Request headers cannot be mutated, so remove any methods that mutate
+export type RequestHeaders = Omit<Headers, "append" | "delete" | "set">;
```
davenewza, line 74: 👍

davenewza (approved):

PR #1305, opened 2023-11-21
feat: sso login authentication flow

davenewza:
# "SSO Login" Authentication Flow

3rd-party "done for you" authentication with our built-in auth providers and any custom OpenID Connect provider.  

In a nutshell:
 1. The provider is configured in `keelconfig.yaml` and the client secret added as an env var to the host.
 2. The `/auth/login/mygoogleclient` endpoint kicks off the authentication process with the provider named "mygoogleclient" in the keelconfig.yaml.
 3. The provider will then request the user authenticate through the necessary means (for example, a username and password page).
 4. Upon successful authentication, the provider will callback to Keel, and the handshake will be complete.
 5. The identity will be created or updated.
 6. The response will be our standard token response (see below).

## Response

The response has an identical structure to the token endpoint.  A Keel `access_token` is returned, which is then used as a bearer token to access Keel actions.  This token can be refreshed using `refresh_token` and the refresh grant.

```json
{
  "access_token": "eyJhbGciOiJSUzI1...",
  "token_type": "bearer",
  "expires_in": 300,
  "refresh_token": "ujnzOSplWfLUW..."
}
```

## Configuration

As with OIDC Token authentication, providers are configured in just the same way in `keelconfig.yaml`.  When configured, a provided will support the SSO Login flow by default.  For example:

```yaml
auth:
  tokens:
    accessTokenExpiry: 30
    refreshTokenExpiry: 60

  providers:
    - type: google
      name: mygoogleclient
      issuerUrl: https://dev-myapp.us.auth0.com/
      clientId: s23tggGssSA5qdsF22gSSdsbwq
```

The client secret must then be added to each host as an *environment variable*.  This can be done either through shell (on local) or with the Console for hosted environments.  The template pattern for the environment variable name is based on the name of the provider:

```
KEEL_AUTH_PROVIDER_SECRET_{providername}
```

For example, `KEEL_AUTH_PROVIDER_SECRET_MYGOOGLECLIENT`

## Still to do

 - REAL WORLD TESTING as there is some uncertainty here
 - Discuss: what do we call this flow?  "OAuth" might be a bit misleading.
 - Discuss: are we happy with the secrets being dynamic env var names?
 - Dynamic `redirect_url` which can be configured from the Backend
 - Remove `oauth` type from keelconfig.yaml - we will only be supporting custom oidc.  We will continue to add on new built-in providers though.
 - On startup, check that all provider secrets exist.

### DIFF
--- a/runtime/apis/authapi/oauth_endpoint.go
+++ b/runtime/apis/authapi/oauth_endpoint.go
@@ -17,39 +17,55 @@ import (
 	"github.com/teamkeel/keel/runtime/common"
 	"github.com/teamkeel/keel/runtime/oauth"
 	"github.com/teamkeel/keel/runtime/runtimectx"
-	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/codes"
-	"go.opentelemetry.io/otel/trace"
 	"golang.org/x/oauth2"
 )
 
-func OAuthHandler(schema *proto.Schema) common.HandlerFunc {
-	return func(r *http.Request) common.Response {
-		return common.Response{
-			Status: http.StatusNotImplemented,
-		}
-	}
-}
+// Error response types for the authorization endpoint
+// https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1
+const (
+	// The request is missing a required parameter, includes an
+	// invalid parameter value, includes a parameter more than
+	// once, or is otherwise malformed.
+	AuthorizationErrInvalidRequest = "invalid_request"
+	// The client is not authorized to request an authorization
+	// code using this method.
+	AuthorizationErrUnauthorizedClient = "unauthorized_client"
+	// The resource owner or authorization server denied the
+	// request.
+	AuthorizationErrAccessDenied = "access_denied"
+	// The authorization server encountered an unexpected
+	// condition that prevented it from fulfilling the request.
+	// (This error code is needed because a 500 Internal Server
+	// Error HTTP status code cannot be returned to the client
+	// via an HTTP redirect.)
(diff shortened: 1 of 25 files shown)


runtime/apis/authapi/oauth_endpoint.go
```diff
@@ -14,3 +30,206 @@ func OAuthHandler(schema *proto.Schema) common.HandlerFunc {
...
+		}
+
+		response := &TokenResponse{
+			AccessToken:  accessTokenRaw,
+			TokenType:    TokenType,
+			ExpiresIn:    int(expiresIn.Seconds()),
+			RefreshToken: refreshToken,
+		}
+
+		return common.NewJsonResponse(http.StatusOK, response, nil)
```
jonbretman, line 213: Shouldn't this callback handler do a redirect back to the initiating app with the token?

davenewza: @jonbretman That is exactly the part I am not sure how to do best.  Right now, if you call on `/auth/login/google` (or whatever), you will be sent to authenticate with that provider, and that will then redirect to the Keel callback handler, which will generate and respond with our tokens:

image

Do you think the best thing would be for our callback to then redirect to some customer-configured web page?  And with the above payload as a redirect header `Location`: `https://myapp.com/callback?access_token={access_token}&token_type={bearer}&refresh_token={refresh_token}...`

jonbretman: Yes exactly, that is how I would expect this to work. The redirect URL would probably be configured in keelconfig so maybe that is another thing we need there.

PR #1307, opened 2023-11-22
chore: using auth code flow

davenewza:
# Using Auth Code Flow

The better thing to do is to redirect from our SSO Login with an authorization code (and not the actual tokens, aka. Implicit Flow, [which is problematic](https://www.taithienbo.com/why-the-implicit-flow-is-no-longer-recommended-for-protecting-a-public-client/)), which can then be traded with access & refresh tokens at the token endpoint - this is standard Authorization Code Flow.  We can also easily extend this with PKCE for customers who need it for their SPA and mobile apps.

## Configuration

In order to use SSO Login, a `redirectUrl` must be configured in `keelconfig.yaml`.

```yaml
auth:
  redirectUrl: https://my-keel-application.com/signedin

  providers:
    - type: google
      name: google
      clientId: 1234567
```

## Example

After authenticating:

image

Using that code to access tokens:

image

## To do

Needs more tests.  The SSO callback also needs to have his error responses reworked so that we are returning errors in the redirect URL query - see https://www.oauth.com/oauth2-servers/authorization/the-authorization-response/

### DIFF
--- a/runtime/apis/authapi/oauth_endpoint.go
+++ b/runtime/apis/authapi/oauth_endpoint.go
@@ -74,13 +74,13 @@ func LoginHandler(schema *proto.Schema) func(http.ResponseWriter, *http.Request)
 		url := oauthConfig.AuthCodeURL(uniuri.New())
 		http.Redirect(w, r, url, http.StatusTemporaryRedirect)
 
-		return common.NewJsonResponse(http.StatusInternalServerError, "oauth redirect failed", nil)
+		return common.NewJsonResponse(http.StatusTemporaryRedirect, "login handler redirect", nil)
 	}
 }
 
 // CallbackHandler is called by the provider after the authentication process is complete
-func CallbackHandler(schema *proto.Schema) common.HandlerFunc {
-	return func(r *http.Request) common.Response {
+func CallbackHandler(schema *proto.Schema) func(http.ResponseWriter, *http.Request) common.Response {
+	return func(w http.ResponseWriter, r *http.Request) common.Response {
 		ctx, span := tracer.Start(r.Context(), "Callback Endpoint")
 		defer span.End()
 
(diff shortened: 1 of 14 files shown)


jonbretman:
I wonder if redirect url should be per provider. Don't you need to know which provider was used?

davenewza:
> I wonder if redirect url should be per provider. Don't you need to know which provider was used?

At that point all you need to deal with are Keel artifacts (Keel auth code to get Keel access and refresh tokens).  The authentication process does not require you to know which provider authenticated the user.  After you've configured `keelconfig.yaml` with providers, the flow is quite abstracted away from these providers.  This doesn't mean we can't include the provider name as a querystring parameter if it would be useful.  What do you think @jonbretman ?

jonbretman:
> > I wonder if redirect url should be per provider. Don't you need to know which provider was used?
> 
> At that point all you need to deal with are Keel artifacts (Keel auth code to get Keel access and refresh tokens). The authentication process does not require you to know which provider authenticated the user. After you've configured `keelconfig.yaml` with providers, the flow is quite abstracted away from these providers. This doesn't mean we can't include the provider name as a querystring parameter if it would be useful. What do you think @jonbretman ?

Makes sense, let's leave it top-level in auth config then 👍

runtime/apis/authapi/oauth_endpoint.go
```diff
@@ -178,27 +178,35 @@ func CallbackHandler(schema *proto.Schema) common.HandlerFunc {
...
+		if err != nil {
+			return common.InternalServerErrorResponse(ctx, err)
 		}
 
-		return common.NewJsonResponse(http.StatusOK, response, nil)
+		values := url.Values{}
+		values.Add("code", authCode)
+		redirectUrl.RawQuery = values.Encode()
+
+		http.Redirect(w, r, redirectUrl.String(), http.StatusFound)
```
jonbretman, line 205: Feels weird to do this when this function returns a `common.Response`. Either we have the `Response` abstraction or we don't. I would just use the `Response` return value to indicate the status and headers.

davenewza: Totally agree - let me fix that up

PR #1313, opened 2023-11-27
feat: mutable id field

davenewza:
# Mutable `id` field

It is now possible to set the value of `id`  during a `create` or an `update` action:

```gql
model Person {
  // fields
  actions {
    create createPerson() with (id, name)
    update updatePerson(id) with (id)
  }
}
```
This is also works in the set attribute, for example: `@set(id = someId)`

NOTE:  When updating `id`, any foreign key columns won't also be updated.  However, I think we can achieve this using   Poastgre's ON UPDATE CASCADE.

## Nested data

This is also supported when creating related model data:

```gql
model Person {
  // fields
  actions {
    create createPerson() with (id, name, employer.id, employer.name)
  }
}
```

### Creating or associating with nested inputs?

This change has an interesting implication.  Previously, it was only possible to either specify the `id` of a nested field (to form an association) or to provide all the desired fields to create the nested model.  This has all changed, because it is now possible to provide `id` in both _associating_ and _creating_ of related data.

**NB: With this change, inputs lead the charge on whether a related model is being _created_ or _associated_.**  

For example, the inputs here imply we are _associating_:

```gql
create createPerson() with (id, name, employer.id)
```

The inputs here imply that we are _creating_ (along with providing our own `id`):

```gql
create createPerson() with (id, name, employer.id, employer.name)
```

Therefore, this will be validated against, because we are trying to set something which isn't in scope of what's being created:

```gql
create createPerson() with (id, name, employer.id, employerName: Text) {
  @set(person.employer.name = employerName)
}
```

This will also fail validation because we are trying to set something which isn't in scope of what's being created:

```gql
create createWithEmployerUsingSetBoth() with (id, name, employerId: ID, employerName: Text) {
  @set(person.employer.id = employerId)
  @set(person.employer.name = employerName)
}
```


jmeistrich:
🥳

PR #1326, opened 2023-12-21
feat: set oneOf json schema property for id and string messages

louisaspicer:
For ID and String messages, we want to set the json schema to use `oneOf` so that only one of the filters can be used at a time. Tom and I had a chat about this and agreed that we would only do this for ID and String fields for now and only allowing one filter at a time makes sense for both string and id filters.

Looking for feedback on how best to approach this

Attempting to go from:
```
"IdQueryInput": {
        "type": "object",
        "properties": {
          "equals": { "type": ["string", "null"] },
          "notEquals": { "type": ["string", "null"] },
          "oneOf": { "type": "array", "items": { "type": "string" } }
        },
        "additionalProperties": false
      }
```
to:
```
"IdQueryInput": {
        "type": "object",
        "oneOf": [
          {
            "title": "Equals",
            "properties": {
              "equals": { "type": "string" }
            },
            "required": ["equals"],
            "additionalProperties": false
          },
          {
            "title": "Not equals",
            "properties": {
              "notEquals": { "type": "string" }
            },
            "required": ["notEquals"],
            "additionalProperties": false
          },
          {
            "title": "One of",
            "properties": {
              "oneOf": { "type": "array", "items": { "type": "string" } }
            },
            "required": ["oneOf"],
            "additionalProperties": false
          }
        ],
        "additionalProperties": false
      }
```

### DIFF
--- a/runtime/jsonschema/testdata/list_on_related_field.json
+++ b/runtime/jsonschema/testdata/list_on_related_field.json
@@ -1,76 +1,118 @@
 {
-	"type": "object",
-	"properties": {
-		"after": {
-			"type": "string"
-		},
-		"before": {
-			"type": "string"
-		},
-		"first": {
-			"type": "number"
-		},
-		"last": {
-			"type": "number"
-		},
-		"where": {
-			"$ref": "#/components/schemas/TestActionWhere"
-		}
-	},
-	"additionalProperties": false,
-	"required": ["where"],
-	"components": {
-		"schemas": {
-			"StringQueryInput": {
-				"type": "object",
-				"properties": {
-					"contains": {
-						"type": "string"
-					},
-					"endsWith": {
-						"type": "string"
-					},
-					"equals": {
-						"type": ["string", "null"]
-					},
-					"notEquals": {
-						"type": ["string", "null"]
-					},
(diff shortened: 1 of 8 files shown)


runtime/jsonschema/testdata/list_on_related_field.json
```diff
@@ -1,76 +1,118 @@
...
+                  "type": "string"
+                },
+                "type": "array"
+              }
+            },
+            "title": "oneOf",
+            "type": "object"
+          }
+        ],
+        "additionalProperties": false
```
jonbretman, line 91: I think you might not want this `additionalProperties` field, as it might be conflicting with `oneOf`.

louisaspicer: Yea that was the issue - Fixed with `unevaluatedProperties` https://json-schema.org/understanding-json-schema/reference/object#unevaluated-properties
Also had to change the tests to allow multiple errors per field.

davenewza (approved):
Looks good!  👍

louisaspicer:
Don't want to merge this yet until we updte the frontend console to handle new format

PR #1344, opened 2024-01-02
fix empty strings being parsed into invalid dates by client js

jmeistrich:
Empty strings are being parsed as invalid dates because the RFC3339 regexp matches on empty strings. So I just changed it to check for truthiness first. You might prefer to adjust the regex instead, but this seems to work well.


davenewza (approved):
Nice. Looks good to go.

davenewza:
You'll just need to use `git amend --commit` to fix your commit message to be `fix: empty strings being...` (notice the colon) for the sake of our [conventional commits policy](https://www.conventionalcommits.org/en/v1.0.0/)

jmeistrich:
Done.

PR #1347, opened 2024-01-05
perf: smarter transactions and authorization for built-in actions

davenewza:
## Performance improvements on built-in actions

Summary:
 - `create`: reduced from 5 to 2 - 5 db hits
 - `update`: reduced from 6 to 2 - 4 db hits
 - `delete`: reduced from 6 to 2 - 4 db hits

### Smarter authorization

For all action types, we perform better early authorization checks, which means we can avoid unnecessary database hits when row-based permissions aren't being used.  For `create`, we even avoid the use of transactions entirely.

Typically, this is the order of database events (i.e. incurring a DB 'hit') for an built-in `create` action:
 a) open transaction
 b) execute create statement
 c) perform row-based permissions query
 d) commit (or rollback) transaction

That is FOUR hits to the database, which is expensive especially when no row-based permissions are even performed.

Now we check if transaction and row-based permission checks are even needed.  If they aren't needed, then the action will be reduced to a single query - executing the insert statement without a transaction block

### No more transactions on `update` and `delete`

Furthermore, transactions were only in place on `update` and `delete` as a fail safe if multiple rows happened to be affected (which would be an error on Keel's part).  We have so many tests and I think the risk of this happening is so low.  The cost of a transaction for every single action call is quite high, so I've opted for dropping it.

### DIFF
--- a/runtime/actions/get.go
+++ b/runtime/actions/get.go
@@ -17,7 +17,7 @@ func Get(scope *Scope, input map[string]any) (map[string]any, error) {
 		return nil, common.NewPermissionError()
 	}
 
-	query := NewQuery(scope.Context, scope.Model)
+	query := NewQuery(scope.Model)
 
 	// Generate the SQL statement
 	statement, err := GenerateGetStatement(query, scope, input)
--- a/runtime/actions/get.go
+++ b/runtime/actions/get.go
@@ -1,8 +1,22 @@
 package actions
 
-import "github.com/teamkeel/keel/runtime/common"
+import (
+	"github.com/teamkeel/keel/proto"
+	"github.com/teamkeel/keel/runtime/common"
+)
 
 func Get(scope *Scope, input map[string]any) (map[string]any, error) {
+	permissions := proto.PermissionsForAction(scope.Schema, scope.Action)
+
+	// Attempt to resolve permissions early; i.e. before row-based database querying.
+	canResolveEarly, authorised, err := TryResolveAuthorisationEarly(scope, permissions)
+	if err != nil {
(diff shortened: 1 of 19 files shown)


runtime/actions/get.go
```diff
@@ -1,9 +1,23 @@
...
 func Get(scope *Scope, input map[string]any) (map[string]any, error) {
-	query := NewQuery(scope.Context, scope.Model)
+	permissions := proto.PermissionsForAction(scope.Schema, scope.Action)
+
+	// Attempt to resolve permissions early; i.e. before row-based database querying.
+	canResolveEarly, authorised, err := TryResolveAuthorisationEarly(scope, permissions)
+	if err != nil {
+		return nil, err
+	}
+	if canResolveEarly && !authorised {
```
louisaspicer, line 16: Is it possible the function returns `authorised` as false but `canResolveEarly` as true so we should allow the query to run?

davenewza: Yeah, that's what is happening here.  Or do you mean the other way around?  `canResolveEarly` as false but and `authorised` as true?

louisaspicer: oh yes sorry i mean if that's the case, is it right that we return a permission error - shouldn't we allow the query to go ahead? obviously ignore me if i'm on the wrong page

louisaspicer (approved):

runtime/actions/get.go
davenewza: Basically if we can deduce early that authorization has failed then there is no point running the query (because auth has failed)

PR #1352, opened 2024-01-09
Group together all equals, notEquals and oneOf message fields

louisaspicer:
This change means messages that are enum types (which also have only equals, notEquals and oneOf fields), will also be grouped together in the jsonschema. Previously we were only grouping `StringQueryInput` and `IdQueryInput`, but enums e.g. `Status` would allow you to set all three queries.


murej (approved):

davenewza:
@louisaspicer LGTM!  Also glad to see you've used `chore: ` as that won't bump the minor version on the next release 👍

PR #1355, opened 2024-01-11
chore: update any type in jsonschema to use anyOf

louisaspicer:
Looking into the OpenAPI docs, our previous implementation of the `any` type was not valid.
Updated to use the `anyOf` property as suggested in the docs.


davenewza (approved):
Looks great.

PR #1356, opened 2024-01-12
feat: javascript client support for 3rd party auth providers

davenewza:
# 3rd-party auth support in the js client

## 💡 Token Abstraction 

Tokens are handled internally by the library.  The client does not need to concern itself with neither access tokens nor refresh tokens.

## ♻️ Automatic Token Refresh 

Token refreshing happens automatically and will also behave correctly with access and refresh tokens of any expiry duration.  It also supports **Refresh Token Rotation**, if enabled by the server.

### Instantiating the client:

In order for users to have long-lived sessions it is necessary to securely store the refresh token either on the server or in a HttpOnly cookie.  When instantiating the client, you can provide your own implementation of `TokenStore`, which is simply made up of a `get` and a `set`.

```ts
const client = new APIClient({
    baseUrl: "https://my-keel-app.com/api",
    {
      get: () => {
        // Get the refresh token (perhaps from cookies)
        return refreshToken;
      },
      set: (token: string) => {
        // Securely set the refresh token 
      }
    }
});
```

NOTE:  if you do not provide your own implementation of `TokenStore`, then the default implementation will hold the refresh token in the client's state, which means automatic refreshing will still work fine but authentication cannot persist across browser refreshes or sessions.

### Get supported authentication providers and their login URLs:

```ts
const providers = await client.auth.providers();
```

The URLs returned here can be used to kick off an SSO flow in the clients browser.  

### Authenticate with an ID token:

```ts
await client.auth.authenticateWithIdToken({ idToken: "xyz" });
```

### Authenticate with SSO:

```ts
await client.auth.authenticateWithSingleSignOn({ code: "xyz" });
```

The token comes from completing the SSO flow.  The 'authorize' URLs to kick off Single Sign On are available from `providers`.

### Check if the session is still authenticated:

```ts
const active = await keel.auth.isAuthenticated();
```

This will return `true` if the access token is still valid.  ALSO, it will refresh the access token implicitly if it is not.  Only if the refresh token has expired (or has been revoked) will it return `false`.

### Logout:

```ts
await  keel.auth.logout();
```

This will also revoke the refresh token on the server.

## 🧪 Tests

I've included some integration tests for the old authenticate action, just to be sure I didn't break anything along the way.


node/client.go
```diff
@@ -405,6 +417,153 @@ class Core {
...
+
+    // Refresh the session with the authentication server.
+    refreshSession: async () => {
+      if (!this.session) {
+        return;
+      }
+
+      const req: RefreshGrant = {
+        grant:  "refresh_token",
+        refreshToken: this.session!.refreshToken
```
TejasQ, line 486: Is this needed since a few lines above we ensure it's defined?

node/client.go
```diff
@@ -405,6 +417,153 @@ class Core {
...
+          };
+          break;
+        case "refresh_token":
+          body = {
+            "grant_type": req.grant,
+            "refresh_token": req.refreshToken
+          };
+          break;
+        default:
+          throw new Error("grant not implemented")
```
TejasQ, line 514: We can do better here.

```suggestion
          throw new Error("Unknown grant type. We currently support `authorization_code`, `token_exchange`, and `refresh_token` grant types. Please use one of those. For more info, please refer to the docs at https://docs.keel.so/authentication/endpoints#parameters")
```

node/client.go
```diff
@@ -405,6 +417,153 @@ class Core {
...
+          cache: "no-cache",
+          headers: {
+            accept: "application/json",
+            "content-type": "application/json",
+          },
+          body: JSON.stringify(body),
+        },
+      );
+     
+      if (result.status == 200) {
```
TejasQ, line 531: Why not

```suggestion
      if (result.ok) {
```

?

node/client.go
```diff
@@ -495,4 +652,46 @@ export type APIError =
...
+
+export interface Session {
+  accessToken: string;
+  refreshToken: string;
+};
+
+export type TokenExchangeGrant = {
+  grant: "token_exchange";
+  subjectToken: string;
+};
```
TejasQ, line 672: Why are we mixing `type` and `interface`? Normally, `type` is better since it cannot be redeclared, whereas `interface`s compose into each other when redeclared.

TejasQ (commented):
This is wonderful. Here are some eyes 👀.

node/client.go
```diff
@@ -405,6 +417,153 @@ class Core {
       }
     },
   };
+
+  auth = {
+    // Manually set the session with an access and refresh token acquired elsewhere.
+    // We recommend to rather use the built-in authentication functions provided in this client.
```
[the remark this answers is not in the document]
adaam2: Needs to use tsdoc syntax for that to work, e.g:

```typescript
/**
 * foo does bar
 */
function foo() {}
```

node/client.go
```diff
@@ -405,6 +417,153 @@ class Core {
...
+            body: JSON.stringify({
+              token: this.session?.refreshToken,
+            }),
+          },
+        );
+  
+        this.session = null;
+        this.client.clearToken();
+    },
+  };
```
adaam2, line 566: A lot of the client doesn't seem like it should be code generated. Might be better off stored in a `.ts` file

node/client.go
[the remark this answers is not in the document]
davenewza: Ive added tsdocs for the auth functions 👍

node/client.go
davenewza: Fixed - thanks

node/client.go
davenewza: Great suggestion - I've pulled all the static code out into a couple of `.ts` files

davenewza:
@TejasQ @adaam2 I've addressed your comments, reworked the API and implemented automatic token refresh.  Would love your 👀

node/templates/client/core.ts
```diff
@@ -0,0 +1,364 @@
...
+        method: "GET",
+        cache: "no-cache",
+        headers: {
+          "content-type": "application/json",
+        },
+      });
+
+      if (result.ok) {
+        const rawJson = await result.text();
+        return JSON.parse(rawJson);
```
adaam2, line 180: Can just use `await result.json()` here too

node/templates/client/core.ts
```diff
@@ -0,0 +1,364 @@
...
+      }
+    },
+  };
+
+  auth = {
+    /**
+     * Returns the list of supported authentication providers and their SSO login URLs.
+     */
+    providers: async (): Promise<Provider[]> => {
+      let url = new URL(this.config.baseUrl);
```
adaam2, line 169: Do we need to use `let` here?

node/templates/client/core.ts
```diff
@@ -0,0 +1,364 @@
...
+          "unexpected status code response from /auth/providers: " +
+            result.status
+        );
+      }
+    },
+
+    /**
+     * Returns true if the session has not expired. If expired, it will attempt to refresh the session from the authentication server.
+     */
+    isAuthenticated: async () => {
```
adaam2, line 192: I would name this something different as it's doing more than just returning a bool

node/templates/client/core.ts
davenewza: Oh yeah, good catch.

node/templates/client/core.ts
davenewza: Fixed

node/templates/client/types.d.ts
```diff
@@ -0,0 +1,125 @@
...
+  headers?: RequestHeaders;
+};
+
+// Result types
+
+export type APIResult<T> = Result<T, APIError>;
+
+type Data<T> = {
+  data: T;
+  error?: never;
```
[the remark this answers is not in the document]
davenewza: Not sure to be honest.  This is existing code.  Will look into it.

PR #1358, opened 2024-01-15
chore: remove additionalProperties for isAny type in jsonschema

louisaspicer:
`additionalProperties` is not required for `any` types in the jsonschema and breaks the RJSF ui on tools as no additional property names are provided:


davenewza (approved):
LGTM

PR #1362, opened 2024-01-16
chore: set title on oneOf properties in the jsonschema

louisaspicer:
This change allows our tools UI to know what title to set per `oneOf` input option.

### DIFF
--- a/runtime/jsonschema/jsonschema.go
+++ b/runtime/jsonschema/jsonschema.go
@@ -344,6 +344,10 @@ func jsonSchemaForField(ctx context.Context, schema *proto.Schema, action *proto
 			message := proto.FindMessage(schema.Messages, m.Value)
 			component := JSONSchemaForMessage(ctx, schema, action, message)
 
+			// Components of oneOf properties should only have one field per property and we should set a title.
+			oneOfFieldName := message.Fields[0].Name
+			component.Title = oneOfFieldName
+
 			// If that nested message component has ref fields itself, then its components must be bundled.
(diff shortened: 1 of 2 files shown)


runtime/jsonschema/jsonschema.go
```diff
@@ -344,6 +344,10 @@ func jsonSchemaForField(ctx context.Context, schema *proto.Schema, action *proto
 			message := proto.FindMessage(schema.Messages, m.Value)
 			component := JSONSchemaForMessage(ctx, schema, action, message)
 
+			// Components of oneOf properties should only have one field per property and we should set a title.
```
davenewza, line 347: Im not sure what this `title` property represents.  Would you mind explaining?

louisaspicer: So a `title` (and there's also `description`) is not required but it provides a way of annotating parts of the schema. 
Here's a bit more info on annotations: https://json-schema.org/understanding-json-schema/reference/annotations

And RJSF (the form library we use for tools) supports using this annotation to set the enum option title for `anyOf` or `oneOf` fields. https://rjsf-team.github.io/react-jsonschema-form/docs/usage/single/#custom-labels-for-enum-fields

If we don't set a title, the enum option just appears as "Option 1" etc.

Do you have a suggestion to make the comment clearer? Thanks!

davenewza: Aah!  I see - makes sense

davenewza (approved):
LGTM!

PR #1370, opened 2024-01-19
feat: default api configurable

davenewza:
## Default API configurable

Currently if no API is specified on the schema we will generate a default API at `/api/` which includes all models and actions.  This change builds on https://github.com/teamkeel/keel/pull/1366 and introduces the following functionality:
 - The default API will be generated (by default) regardless of any additional API defined in the schema
 - `useDefaultApi` config value in `keelconfig.yaml` which lets you enable or disable the default API - `useDefaultApi: false` will disable it
 - If "Api" is already defined in the schema, then it will override the generation of the default API


schema/schema.go
```diff
@@ -50,20 +50,30 @@ func (scm *Builder) MakeFromDirectory(directory string) (*proto.Schema, error) {
...
+	scm.Config = config
+
 	return scm.makeFromInputs(&reader.Inputs{
 		SchemaFiles: scm.schemaFiles,
 	})
 }
 
 // MakeFromFile constructs a proto.Schema from the given inputs
 func (scm *Builder) MakeFromInputs(inputs *reader.Inputs) (*proto.Schema, error) {
+	scm.Config = &config.ProjectConfig{}
```
louisaspicer, line 73: What's the reason for an empty config here?

louisaspicer (approved):

schema/schema.go
davenewza: I think it's better that, when not set, the config is empty rather than nil.  This means we don't need to nil check everywhere in the runtime

Issue #1371, opened 2024-01-20
Bug: Errors in test lack context


xakdog:
## Problem

Errors in tests provide little to no context. Here's an example

<img width="889" alt="image" src="https://github.com/teamkeel/keel/assets/17473315/72fec5d8-52d7-411d-9597-9c5f97b88ef4">

## How to reproduce

Generate a default project and add `tests/main.test.ts` file

```ts
import { expect, test } from 'vitest';
import { actions, models } from '@teamkeel/testing';
 
test('a problematic test', async () => {
  const johnDoe = await models.identity.create({
    id: 'john', // <-- this is the problem
    email: 'john@example.com',
  });

  const project = await models.project.create({
    title: 'Rust Doge',
    ownerId: johnDoe.id,
  });

  const fetchedRecord = await actions.withIdentity(johnDoe).getProject({ id: project.id });

  expect(fetchedRecord).toEqual(project);
});
```

davenewza:
The underlying issue here is that we recently added support for setting a custom, non-KSUID, value for the `id` column, and clearly this was missed during development.  I've put together a fix and tests for this issue here #1373.  Great find and thank you for reporting it.

Typically a `ERR_INTERNAL` should never happen in production.  It means "something went very wrong".  I've approved #1372 to help us with troubleshooting these issues.

PR #1372, opened 2024-01-20
fix: propagate error messages for httpjson in test environment

xakdog:
## Problem

Errors in tests lack context. Check the issue https://github.com/teamkeel/keel/issues/1371

image

## Solution

Propagate errors in a test environment

image


davenewza (approved):
See https://github.com/teamkeel/keel/pull/1373 for the fix for the underlying issue

I'm happy to include your change as it will help us troubleshoot issues quicker while we're leading up to a Keel v1.  Thank you for your contribution @xakdog

PR #1374, opened 2024-01-22
fix: notequals for id in modelapi

davenewza:
### notEquals for `id` field on the ModelAPI

Fix for issue reported here: https://discord.com/channels/1135925644098809907/1135931890621239326/1198899132283236366


davenewza:
https://linear.app/keel/issue/KE-1124/notequals-for-idwherecondition-in-model-api

PR #1375, opened 2024-01-22
fix: add console auth provider on cli

davenewza:
### Local Console auth token fix

When using the local Console we are getting failed token requests because the Console's auth provider has not been configured in the customer's keelconfig.yaml.  We therefore add this to the configuration.  Note that we are only adding it for ID Token authentication and therefore no client secret is required.

### DIFF
--- a/cmd/program/model.go
+++ b/cmd/program/model.go
@@ -58,6 +58,11 @@ const (
 	StatusQuitting
 )
 
+const (
+	consoleAuthProviderIssuer   = "https://auth.keel.xyz/"
+	consoleAuthProviderClientId = "KvnOmNPy17WtMGBseDZRw3Hgn0ZOQpDd"
+)
+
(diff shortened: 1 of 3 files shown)


cmd/program/model.go
```diff
@@ -58,6 +58,11 @@ const (
 	StatusQuitting
 )
 
+const (
+	consoleAuthProviderIssuer   = "https://auth.keel.xyz/"
+	consoleAuthProviderClientId = "KvnOmNPy17WtMGBseDZRw3Hgn0ZOQpDd"
```
louisaspicer, line 63: Maybe silly question but is it ok to just have the clientId in public view?

davenewza: It's a good question.  Yeah, it's harmless.  https://devforum.okta.com/t/is-clientid-for-an-app-confidential/5841.  It's the client secret which is sensitive

louisaspicer (approved):

PR #1416, opened 2024-02-06
fix: client token refresh improvements

davenewza:
## Improvements to the client auth

This is based on ongoing work with @TejasQ 

 - Added password flow auth
 - Refresh and access tokens are not storable.
 - Changed the API to get and set tokens.
 - Improved the refresh flow
 - Added some much needed tests

Not done is returning the identity when authenticating.  This will probably require us either to create a OIDC user-info endpoint or to return an ID token from the token endpoint with all the necessary user claims, which we can then parse properly in the client.

### DIFF
--- a/node/templates/client/core.ts
+++ b/node/templates/client/core.ts
@@ -1,12 +1,9 @@
 export class Core {
   constructor(
     private config: RequestConfig,
-    private refreshToken: TokenStore = new InMemoryTokenStore()
-  ) {
-    this.auth.refresh();
-  }
-
-  #session: AccessTokenSession | null = null;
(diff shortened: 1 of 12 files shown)


node/templates/client/core.ts
```diff
@@ -186,19 +183,46 @@ export class Core {
...
+        return new Date(exp * 1000);
+      }
+
+      throw new Error("jwt token could not be parsed");
+    },
+
     /**
      * Returns true if the session has not expired. If expired, it will attempt to refresh the session from the authentication server.
      */
     isAuthenticated: async () => {
```
davenewza, line 213: Adam suggested we might rename this, because it actually might cause a request to the refresh the token

PR #1435, opened 2024-02-14
fix: deprecate authenticate

davenewza:
## Deprecate the `authenticate()` action

The built-in action is deprecated.  User/password authentication must use the password grant on the `/auth/token` endpoint.


louisaspicer (approved):
🙌

PR #1436, opened 2024-02-16
feat: generate client auth support

davenewza:
# 3rd-party auth support in the generated client

## 💡 Token Abstraction 

Tokens are handled internally by the library.  The client does not need to concern itself with neither access tokens nor refresh tokens.

## ♻️ Automatic Token Refresh 

Token refreshing happens automatically and will also behave correctly with access and refresh tokens of any expiry duration.  It also supports **Refresh Token Rotation**, if enabled by the server.

### Instantiating the client:

By default, we store persistence in memory, which means that reinstantiating the client will require the user to reauthenticate.

In order for users to have long-lived sessions it is necessary to securely store the refresh token either on the server or in a HttpOnly cookie.  When instantiating the client, you can provide your own implementation of `TokenStore`, which is simply made up of a `get` and a `set`.

```ts
class LocalStorageStore {
  #key = "keel-refresh-token";

  get = () => {
    return localStorage.get(#key);
  };

  set = (token: string | null): void => {
    if (token) {
      localStorage.set(#key, token);
    } else {
      localStorage.removeItem(#key);
    }
  };
}
```

And then configuring the the client as following:

```ts
const client = new APIClient({
  baseUrl: process.env.API_BASE_URL,
  refreshToken: new LocalStorageStore()
});
```

Note that you can also provide a custom store implement for access tokens, however it's important to remember that access tokens are typically short-lived and only need to be stored if you're unable to reuse them between requests.

```ts
const client = new APIClient({
  baseUrl: process.env.API_BASE_URL,
  accessToken: new LocalStorageStore(),
  refreshToken: new HttpOnlyCookieStore()
});
```

### Get supported authentication providers and their login URLs:

```ts
const providers = await client.auth.providers();
```

The URLs returned here can be used to kick off an SSO flow in the clients browser.  

### Authenticate with email and password:

```ts
await client.auth.authenticateWithPassword(email, password);
```

### Authenticate with an ID token:

```ts
await client.auth.authenticateWithIdToken(idToken);
```

### Authenticate with SSO:

```ts
await client.auth.authenticateWithSingleSignOn(authCode);
```

The token comes from completing the SSO flow.  The 'authorize' URLs to kick off Single Sign On are available from `providers`.

### Check if the session is still authenticated:

```ts
const active = await keel.auth.isAuthenticated();
```

This will return `true` if the access token is still valid.  ALSO, it will refresh the access token implicitly if it is not.  Only if the refresh token has expired (or has been revoked) will it return `false`.

### Logout:

```ts
await  keel.auth.logout();
```

This will also revoke the refresh token on the server.

## 🧪 Tests

I've implemented a bunch of tests using our integration test runner.


node/templates/client/core.ts
```diff
@@ -0,0 +1,367 @@
...
+      }
+      return this;
+    },
+    setBaseUrl: (value: string): Core => {
+      this.config.baseUrl = value;
+      return this;
+    },
+    rawRequest: async <T>(action: string, body: any): Promise<APIResult<T>> => {
+      // If necessary, refresh the expired session before calling the action
+      await this.auth.isAuthenticated();
```
louisaspicer, line 24: If we don't need to check isAuthenticated is true/false, could this just be this.auth.refresh() instead?

louisaspicer (approved):

node/templates/client/core.ts
davenewza: `await this.auth.isAuthenticated();` will only refresh the token if the access token is expired, whereas `this.auth.refresh()` wil force a refresh, which would be a waste of time

PR #1437, opened 2024-02-16
chore: update auth openapi definition for token request

louisaspicer:
This should fix the issue we have when translating the openapi auth schema to a form for authenticating in the API explorer.


davenewza (approved):
Looks good!

davenewza:
Let me know if you'd like this on a `next` prerelease

PR #1439, opened 2024-02-19
fix: add BooleanQueryInput to oneOf conditions and create anyOf grouping

louisaspicer:
Updates the remaining input query messages using either `oneOf` or `anyOf` options in the open api schema.
- `BooleanQueryInput` message is now set to use `oneOf`
- `DateQueryInput`, `TimestampQueryInput` and `IntQueryInput` now use `anyOf` - this is so that we can allow group these inputs together in the Console Tools UI, but also allow more than one of their filters to be set.


davenewza (approved):
Nice!!

By the way, prefixing the squashed commit with "chore" will mean it will not increment the semantic version, which means the change won't be released until the next "feat" or "fix" comes along.  So, could make this "fix" instead.

Tomorrow I will also show you a bit about how to do prereleases.  It's super simple

PR #1445, opened 2024-02-22
fix: allow additionalProperties on anyOf options

louisaspicer:
Currently `anyOf` options are set to have `additionalProperties: false` which stops the client from sending a request with multiple filters.


davenewza (approved):
LGTM!

PR #1447, opened 2024-02-23
fix: update auth openapi schema with default value

louisaspicer:
This change is required for when we use the open api schema, we can hide the `grant_type` input field from showing in the UI, as a default value (the const) is already set.


davenewza (approved):
Looks good!

PR #1448, opened 2024-03-04
feat: registration and login flows with identity with password and id token auth

davenewza:
# Registration and login flows with identity with Password and ID Token authentication

### `create_if_not_exists` request parameter

By setting the `create_if_not_exists` to `false` on the token endpoint, the identity will not be created if it does not exist yet.  

This parameter is optional.  The default is `true`.

This is supported for the Password and ID Token flows.

### `identity_created` response field

The token endpoint now includes the field `identity_created` in the response body which informs whether an identity was created or not during the auth operation.

### DIFF
--- a/runtime/apis/authapi/token_endpoint.go
+++ b/runtime/apis/authapi/token_endpoint.go
@@ -165,7 +165,7 @@ func TokenEndpointHandler(schema *proto.Schema) common.HandlerFunc {
 
 			if identity == nil {
 				if !createIfNotExists {
-					return jsonErrResponse(ctx, http.StatusUnauthorized, TokenErrInvalidClient, "possible causes may be that the identity does not exist or the credentials are incorrect", nil)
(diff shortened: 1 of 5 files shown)


runtime/apis/authapi/token_endpoint.go
```diff
@@ -152,6 +164,10 @@ func TokenEndpointHandler(schema *proto.Schema) common.HandlerFunc {
 			}
 
 			if identity == nil {
+				if !createIfNotExists {
+					return jsonErrResponse(ctx, http.StatusUnauthorized, TokenErrInvalidClient, "possible causes may be that the identity does not exist or the credentials are incorrect", nil)
```
tomfrew, line 168: Let's remove "possible causes may be that " from these messages. So it's just "the identity does not exist or the credentials are incorrect".

Same with the other ones in this file

tomfrew (approved):

runtime/apis/authapi/token_endpoint.go
davenewza: Good call.  Fixed.

PR #1451, opened 2024-03-21
fix: split numbers when converting between camelCase and snake_case

RutZap:
The Keel go runtime is snake casing the model field `test1` correctly to `test_1`, which is then also used in migrations, which means that the column in the database will be called `test_1`.  However, when using the ModelAPI in functions, the JavaScript runtime is snake casing the field `test1` to `test1`.  This is incorrect as the ModelAPI is not mapping to the database column correctly. This PR addresses this issue by treating numbers as _words_.

### DIFF
--- a/packages/functions-runtime/src/casing.test.js
+++ b/packages/functions-runtime/src/casing.test.js
@@ -1,5 +1,5 @@
 import { test, expect } from "vitest";
-import { camelCaseObject } from "./casing";
+import { camelCaseObject, snakeCaseObject } from "./casing";
 
 const cases = {
   FROM_SNAKE: {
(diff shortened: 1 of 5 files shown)


packages/functions-runtime/src/casing.test.js
```diff
@@ -1,24 +1,55 @@
...
+      test_1_test: "test",
+      test_20: "test",
     },
   },
 };
 
 Object.entries(cases).map(([key, { input, expected }]) => {
   test(key, () => {
-    const result = camelCaseObject(input);
+    let result;
```
louisaspicer, line 47: You could also write this like:
`const result = key === "FROM_SNAKE" ? camelCaseObject(input) : snakeCaseObject(input)`

But I don't mind either way - longer version is usually clearer anyway 😄

louisaspicer (approved):

packages/functions-runtime/src/casing.test.js
```diff
@@ -1,24 +1,55 @@
...
+      test_1: "test",
+      test_a_1: "test",
+      test_1_test: "test",
+      test_20: "test",
+      test_url: "test",
     },
   },
 };
 
 Object.entries(cases).map(([key, { input, expected }]) => {
```
davenewza, line 49: Not sure why this needed to be a map for only two cases, but it works!

davenewza (approved):
LGTM!

PR #1459, opened 2024-03-25
feat: add new Markdown scalar type

RutZap:
Introduce a new built-in field type: `Markdown`. Its behaviour is consistent with `Text` but by differentiating it from `Text` allows api consumers to make use of rich text editors.


schema/validation/rules/expression/expression.go
```diff
@@ -286,6 +286,7 @@ func OperandTypesMatchRule(asts []*parser.AST, condition *parser.Condition, cont
 	// Possibly this only applies to Date and Timestamp
 	comparable := [][]string{
 		{parser.FieldTypeDate, parser.FieldTypeDatetime},
+		{parser.FieldTypeMarkdown, parser.FieldTypeText},
```
davenewza, line 289: Nice

davenewza (approved):
Looks great.  That's a lot of tests!

integration/testdata/built_in_actions/schema.keel
```diff
@@ -4,12 +4,13 @@ model Post {
...
 
     actions {
-        create createPost() with (title, subTitle)
+        create createPost() with (title, subTitle, content?)
         get getPost(id)
-        update updatePost(id) with (title)
+        update updatePost(id) with (title, content?)
         delete deletePost(id)
         delete deletePostBySubTitle(subTitle)
         list listPosts(id?, title?, category?, rating?)
```
davenewza, line 16: I would have probably added one implicit filtering test and one expression filtering test with `list`

runtime/actions/query_test.go
```diff
@@ -510,12 +512,14 @@ var testCases = []testCase{
...
 					age Number
 					isActive Boolean
+					bio Markdown
 				}
 				actions {
 					update updatePerson(id) {
 						@set(person.name = "Bob")
 						@set(person.age = 100)
 						@set(person.isActive = true)
+						@set(person.bio = "## Biography")
```
adaam2, line 522: What about a test for markdown contain quotes here?

runtime/jsonschema/jsonschema.go
```diff
@@ -395,7 +395,7 @@ func jsonSchemaForField(ctx context.Context, schema *proto.Schema, action *proto
 			prop = JSONSchema{OneOf: oneOf}
 		}
 
-	case proto.Type_TYPE_ID, proto.Type_TYPE_STRING:
+	case proto.Type_TYPE_ID, proto.Type_TYPE_STRING, proto.Type_TYPE_MARKDOWN:
```
louisaspicer, line 398: Btw this is where we should change it to another case and have 
```
    prop.Type = "string"
    prop.Format = "markdown"
```

PR #1462, opened 2024-03-28
chore: enforce correct number of args for each command

RutZap:
Consolidate CLI commands arguments validation by enforcing the required number of args; e.g:

```
user@pc> keel secrets remove 
Error: accepts 1 arg(s), received 0
Usage:
  keel secrets remove <key> [flags]
...
```

```
user@pc> keel run project  -d ./../starter
Error: unknown command "project" for "keel run"
Usage:
  keel run [flags]
...
```

```
user@pc> keel secrets set TEST "value" dev
Error: accepts 2 arg(s), received 3
Usage:
  keel secrets set <key> <value> [flags]
```

### DIFF
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -1,8 +1,6 @@
 package cmd
 
 import (
-	"os"
-
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/spf13/cobra"
 	"github.com/teamkeel/keel/cmd/program"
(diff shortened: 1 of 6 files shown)


davenewza (approved):
Looks good!  I assume you've tested it quite well (since we don't have automated testing for the CLI).  Please try to write something in the PR description especially since this is a public repo - thanks

PR #1467, opened 2024-04-10
feat: array fields

davenewza:
# Array fields

NOTE:  this is being merged into the `next` branch.  It will not be released just yet as there is still work remaining.

A model field can now be defined as an array - an ordered list of non-unique values.  For example, a blog post can have a number of tags associated to it:

```
model Post {
  fields {
    tags Text[]
  }
}
```

See https://github.com/teamkeel/keel/pull/1469 for more info on `all` and `any` queries (merged into here)

## What is supported?

✅ Arrays for all scalar types (`Text`, `Number`, `Bool`, `Timestamp`, `Date`, and enums)
✅ Arrays can defined as nullable, i.e. `tags Text[]?`
✅ Action input support arrays inputs.
✅ `list` action supports `equals` and `notEquals` for array fields, which will compare the entire array data with that of the array input.
✅ Expressions support array literals, for example `["science", "technology"]` or even the empty array `[]`.
✅ Conditional expressions support `==`, `!=` and `in` for array fields.
✅ Array fields can be given a default value, for example, `@default(["all"])`.
✅ Array fields can be set with expressions, for example, `@set(post.tags = [])`.
✅ Nested `create` inputs for relationships support array inputs.
✅ Arrays correctly parsed to TypeScript arrays for functions and in the test SDK.
✅ `list` action support for `all` and `any` which provides query options to search within arrays.
✅ `ModelAPI` querying for arrays. with `equals`, `notEquals`, `all`, and `any` supported on the `findMany` function. 

## What is currently incomplete?

❌ Using `in` on a array field in a nested to-many relationship, for example `@where("science" in author.books.tags)`


RutZap (approved):

davenewza:
Rather going to merge this straight into main.  See https://github.com/teamkeel/keel/pull/1475

PR #1469, opened 2024-04-11
feat: array any and all queries

davenewza:
# `any` and `all` array queries

With the `list` action, it is now possible to perform queries within array data using the `all` and `any` query operators.  For example, the following query returns all posts **where any tag is equal to "science"**:

```gql
query {
  listPosts(input: {
    where: {
      tags: {
        any: {
          equals: "science"
        }
      }
    }
  }) {
  ...
  }
}
```

All of the scalar query operations are available, with the exception of `contains`, `startsWith` and `endsWith` for Text.

It is also possible to combine queries in just the same way with our existing list querying.  For example:

```gql
query {
  listPosts(input: {
    where: {
      tags: {
        any: {
          equals: "science",
          notEquals: "education"
        },
        all: {
          notEquals: "science"
        }
      }
    }
  }) {
  ...
  }
}
```

This will be generated into a single SQL query.  And, of course, you can combine this with expressions, etc.

Note:  more tests need to be written.

### DIFF
--- a/runtime/actions/list.go
+++ b/runtime/actions/list.go
@@ -56,26 +56,59 @@ func (query *QueryBuilder) applyImplicitFiltersFromMessage(scope *Scope, message
 		}
 
 		valueMap, ok := value.(map[string]any)
-
-		// Cannot be parsed into map
 		if !ok {
 			return fmt.Errorf("'%s' input value %v is not in correct format", input.Name, value)
 		}
 
 		for operatorStr, operand := range valueMap {
-			operator, err := graphQlOperatorToActionOperator(operatorStr)
-			if err != nil {
-				return err
+			var operator ActionOperator
+			var err error
+
+			switch operatorStr {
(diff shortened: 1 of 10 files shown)


runtime/actions/list.go
```diff
@@ -56,26 +56,59 @@ func (query *QueryBuilder) applyImplicitFiltersFromMessage(scope *Scope, message
...
 		for operatorStr, operand := range valueMap {
-			operator, err := graphQlOperatorToActionOperator(operatorStr)
-			if err != nil {
-				return err
-			}
+			var operator ActionOperator
+			var err error
+
+			switch operatorStr {
+			case "any", "all":
```
RutZap, line 68: This is probably not for now but I feel that these should be defined as constants.

davenewza: Good spot. I will change this along with the others in makeproto

PR #1470, opened 2024-04-12
feat: enforce maximum identifier length through validation rules

RutZap:
Postgresql has a maximum identifier length limit of 63 bytes. Anything over 63 bytes will be automatically truncated and this will cause issues with migrations. 

This PR adds validation rules to check that field names and model names will not create identifiers longer than the limit enforced by the DB engine. The limit is applied on the snake cased version of the field/model name, as this is what will be used in order to generate the db identifiers. The limit also takes in account the longest suffix added to generated triggers (e,g, `model_name_updated_at`).

### DIFF
--- a/schema/validation/rules/field/field.go
+++ b/schema/validation/rules/field/field.go
@@ -82,13 +82,13 @@ func ValidFieldTypesRule(asts []*parser.AST) (errs errorhandling.ValidationError
 
 // FieldNamesMaxLengthRule will validate that field names are smaller than the maximum allowed by postgres (63 bytes).
 func FieldNamesMaxLengthRule(asts []*parser.AST) (errs errorhandling.ValidationErrors) {
-	const MAX_BYTES = 63
+	const maxBytes = 63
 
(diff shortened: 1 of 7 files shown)


schema/validation/rules/field/field.go
```diff
@@ -78,3 +79,25 @@ func ValidFieldTypesRule(asts []*parser.AST) (errs errorhandling.ValidationError
 
 	return
 }
+
+// FieldNamesMaxLengthRule will validate that field names are smaller than the maximum allowed by postgres (63 bytes).
+func FieldNamesMaxLengthRule(asts []*parser.AST) (errs errorhandling.ValidationErrors) {
+	const MAX_BYTES = 63
```
davenewza, line 85: We have been using Pascal case for exported consts and camel case for internal consts.  I might be wrong, but I think the standard library uses Pascal and camel as well.  For consistency, I think we should change these.  Unless you feel quite strongly about using C-style?

davenewza (approved):
Looks great

PR #1472, opened 2024-04-19
fix: snake case fields when creating via ModelAPI

RutZap:
When creating models via ModelAPI, all fields need to be passed through the snake case object helper func. This will transform fields such as `address1` into the expected db column `address_1`.


davenewza (approved):
Looks good!

PR #1486, opened 2024-04-30
fix: vector type light

davenewza:
A very incomplete implementation of the `Vector` data type, which stores 1536-length vectors using pgvector.

What's working?
 - `create`, `get`, `list` and `update` using the built-in actions.
 - Vectors fields hidden from GraphQL responses
 - Code completions
 - Validation on vector size for creating and updating

What's not working?
 - ModelAPI
 - Vectors fields hidden from RPC and JSON responses
 - Vector querying operations
 - `Vector[]` should be validated against
 - Console support


migrations/sql.go
```diff
@@ -28,6 +28,7 @@ var PostgresFieldTypes map[proto.Type]string = map[proto.Type]string{
 	proto.Type_TYPE_SECRET:    "TEXT",
 	proto.Type_TYPE_PASSWORD:  "TEXT",
 	proto.Type_TYPE_MARKDOWN:  "TEXT",
+	proto.Type_TYPE_VECTOR:    "VECTOR(1536)",
```
RutZap, line 31:
```suggestion
	proto.Type_TYPE_VECTOR:    "VECTOR",
```

I think passing the size will prevent the field from being used with anything other than 1536 dimensions in this case!?

RutZap (approved):

PR #1489, opened 2024-05-08
fix: check for crypto before importing module

RutZap:
Check if `crypto` is set before attempting to set the property. 

This issue came up when using newer versions of node. It seems that starting with node@19, crypto was no longer experimental and therefore included as a built-in module, `globalThis.crypto`, thus attempting to set it on `globalThis` as a polyfill would end up with an error:

```
TypeError: Cannot set property crypto of #<Object> which has only a getter
```


davenewza (approved):
LGTM but I think it might be good to get @tomfrew's eyes on it.

PR #1492, opened 2024-05-22
Feat: validation improvements & WASM update

RutZap:
Validation errors will now include warnings. These are suppressed by default when using the CLI but in the wasm package an option is available which will include warnings in its output. The WASM validation will now also just validate rather than build the proto schema. This will inevitably improve performance, but it is unknown by how much at this point in time.


davenewza (approved):
Looks great

PR #1497, opened 2024-06-19
fix: audit 'event_processed_at' column cannot be parsed or is empty

davenewza:
It is possible for the event_processed_at column on the `keel_audit` table to not be set when parsing audit logs.  It could be unset due to a legitimate race condition or if an event was not set up for that op type at that time.

We don't need to even parse this column anyway, so I am removing it.


davenewza:
https://linear.app/keel/issue/KE-2044/audit-event-processed-at-column-cannot-be-parsed-or-is-empty

PR #1508, opened 2024-07-03
feat: inline files

RutZap:
Introducing a new field type: `InlineFile`. These fields are files that are being provided as part of the API requests in a data-url format; e.g.  `data:application/pdf;name=MyUploadedFile.pdf;base64,xxxxxx[...]`

By default, when in a CLI context, files are stored locally within the Keel app's database in a new table `keel_storage`.  This PR introduces a `Storage` service which can be injected in the runtime context to switch to a different storage solution (e.g. S3).

The rough process for handling files is:
* When an auto-action (`create` or `update`) that contains a `InlineFile` input is handled, the storage service will be asked to store the file and provide a `storage.FileInfo` struct. This is then Json encoded and stored in the db alongside the affected model (as jsonb).
* When the results are returned to the consumer, the runtime will go through and change each `InlineFile` field into a rich structure which contains file information (size, URL, filename, unique id, etc). At this point, the information is hydrated by calling the storage service (int he event of a S3 implementation for example, a signed URL can be injected)

Example schema
```
model Blog {
    fields {
        title Text
        image InlineFile
        category Category?
    }

    actions {
        get getBlog(id) {
            @embed(category)
        }
        update updateBlog(id) with (title, image?)
        create createBlog() with (title, image, category.id?)
        list listBlogs() {
            @embed(category)
        }
    }

    @permission(
        actions: [get, create, list, update],
        expression: true
    )
}

model Category {
    fields {
        name Text
        blogs Blog[]
        background InlineFile
    }

    actions {
        get getCategory(id)
        create createCategory() with (name, background)
        list listCategories() {
            @embed(blogs)
        }
    }

    @permission(
        actions: [get, create, list, update],
        expression: true
    )
}
``` 
Screenshot 2024-07-03 at 15 35 36


davenewza (approved):
Looks great.  Yeah, I would definitely suggest we write some more tests, but we may need better functions world support first.

PR #1511, opened 2024-07-05
feat: inline files input type support in functions runtime

RutZap:
`InlineFile` inputs are given to the runtime as a string. The open api schema treats any `InlineFile` as a `string`. When we receive a API request to a action with an `InlineFile` input (given as a dataurl) we do this:

* in the case of an auto action: we store the file, get a `FileData` object back and store that in the database.
* in the case of custom functions, after validation, we decorate the input from a simple string to :

```
{
   "field": {
      "__typename": "InlineFile",
      "dataURL": "data:......."
   }
}
```

this is then passed as request params to the custom function lambda. The JS sdk will parse inputs and when it sees `"__typename": "InlineFile"` it will replace the object with a instance of the `sdk.InlineFile` class.

This then means that we can send a normal json http request like:

```
{
  "image": "data:....."
}
```
and in our custom function we can use the `image` input field as an instance of `sdk.InlineFile`:

```
export default CustomFunction(async (ctx, inputs) => {
  console.log(inputs.image.contentType);
  console.log(inputs.image.size)
  console.log(inputs.image.read()); // this returns a Blob at the moment
});
```


davenewza (approved):

davenewza (commented):
Mistakenly approved this.  Changing to "comment"

davenewza (changes requested):
Mistakenly approved this. Changing to "request changes"

integration/testdata/jobs_basic/tests.test.ts
```diff
@@ -92,5 +94,6 @@ test("jobs - all types as input values", async () => {
     timestamp: new Date(2022, 12, 25, 1, 3, 4),
     id: "123",
     enum: Status.GoldPost,
+    image: InlineFile.fromDataURL(dataUrl),
```
davenewza, line 97: nice

packages/functions-runtime/src/InlineFile.js
```diff
@@ -0,0 +1,78 @@
+class InlineFile {
+  constructor(key, filename, contentType, size, url) {
+    this.filename = filename;
+    this.contentType = contentType;
+    this.size = size;
+    this.key = key;
+    this.public = false;
+    this.url = url;
```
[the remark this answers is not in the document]
davenewza: Also, would it be `url` or would it actually be an S3 path?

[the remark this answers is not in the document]
davenewza: I suppose we can figure this out when doing ModelAPI side of things

packages/functions-runtime/src/ModelAPI.test.js
```diff
@@ -61,15 +65,19 @@ beforeEach(async () => {
...
     name: "Jim",
     married: false,
     favouriteNumber: 10,
+    avatar: file,
   });
   expect(row.name).toEqual("Jim");
   expect(row.married).toEqual(false);
   expect(row.favouriteNumber).toEqual(10);
+  expect(row.avatar.filename).toEqual("my-avatar.png");
+  expect(row.avatar.size).toEqual(2024);
```
davenewza, line 80: Is it worth testing the value of `row.file.read()`?

packages/functions-runtime/src/parsing.js
```diff
@@ -0,0 +1,28 @@
+const { InlineFile } = require("./InlineFile");
+
+// parseParams takes a set of inputs and creates objects for the ones that are of a complex type.
+//
+// inputs that are objects and contain a "__typename" field are resolved to instances of the complex type
+// they represent. At the moment, the only supported type is `InlineFile`
+function parseParams(inputs) {
```
davenewza, line 7: This is great!  👏👏

runtime/actions/parsing.go
```diff
@@ -176,3 +177,35 @@ var toDate = func(value any) (types.Date, error) {
 		return types.Date{}, fmt.Errorf("incompatible type %T parsing to Date", t)
 	}
 }
+
+// TransformCustomFunctionsInputTypes will, similarly to TransformInputTypes traverse through the input data structure
+// and will decorate complex input fields with typenames to be used by the JS environment
+//
+// e.g. for InlineFile inputs, whcih are given as a dataURL string, they need to be transformed into an object
+// including the typename
+func TransformCustomFunctionsInputTypes(schema *proto.Schema, action *proto.Action, input any) (any, error) {
```
davenewza, line 186: This isn't recursive, which means it will only work for fields at the root level, right?  It's possible that there could be a nested input data structure using Messages

davenewza (approved):
Looks great!!  

What's missing is tests for event subscriptions, assuming that will work out of the box.  If not, let's create a ticket (and do it soon in my opinion)

I think the PR title is wrong though as this is inline files function inputs?

packages/functions-runtime/src/InlineFile.js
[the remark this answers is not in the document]
RutZap: I'm seeing this as being a S3 path that would be used in `read()` if the dataUrl is not set

packages/functions-runtime/src/ModelAPI.test.js
RutZap: So in the modelAPi.. the `InlineFile` field is not really an `InlineFile` instance yet. When reading via kysely we're getting the object that's been stored in the db. We need to add a hook of some sort to turn it into a InlineFile instance, but this would be part of the following piece of work which involves modelAPI support for _rich_ input types

davenewza: Ah yes, good point!

runtime/actions/parsing.go
RutZap: Agreed. We'll tackle this in a separate PR.

PR #1524, opened 2024-07-16
fix: add completions for embed attributes

RutZap:
Adding completions for `@embed` attributes. The completions will be reduced to a set of model fields (relationships) available on the given model.


davenewza (approved):
LGTM

PR #1526, opened 2024-07-22
feat: identity hooks

davenewza:
# Identity Hooks

It is now possible to provide custom functionality to parts of the authentication lifecycle.  This PR introduces `afterAuthentication` and `afterIdentityCreated` hooks.

The following addition to `keelconfig.yaml` will enable these hooks in the runtime:

```yaml
auth:
  hooks: 
    - afterAuthentication
    - afterIdentityCreated
```

Any invalid hook name will result in an error:

image

As with other functions, you will first need to create them, otherwise

image

**You can run `keel generate` to scaffold these functions!**

Each function has full access to the standard `ctx` object that is passed through with normal custom functions.

Example implementation of `afterIdentityCreated` to create an associated `Account` model:

```ts
export default AfterIdentityCreated(async (ctx) => {
    const identity = ctx.identity!;

    const account = await models.account.findOne({ identityId: identity.id});
    if (!account) {
        await models.account.create({ name: identity.name ?? "Not set", identityId: identity.id });
        console.log("account created")
    } else {
        console.log("account already exists")
    }
});
```

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -8,7 +8,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v3
         with:
-          go-version: "1.20"
+          go-version: "1.21"
       - uses: actions/checkout@v4
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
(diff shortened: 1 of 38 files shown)


davenewza:
Closing this as it was merged into `next` instead.

PR #1530, opened 2024-07-24
feat: remove foreign key from schema when object is embedded

RutZap:
When embedding a related object, we want to remove the foreign key; i.e.:

```
    actions {
        list listBooks() {
            @embed(author)
        }
    }
```

```
 "results": [
    {
      "author": {
        "createdAt": "2024-06-24T12:49:02.765243Z",
        "firstName": "Radu",
        "id": "2iKJTcMSCRsnNA1ifJSQkZ7Kl8d",
        "surname": "Gruia",
        "updatedAt": "2024-06-24T12:49:02.765243Z"
      },
      "authorId": "2iKJTcMSCRsnNA1ifJSQkZ7Kl8d",   <=== This should be removed from responses
      "createdAt": "2024-06-24T12:49:13.964459Z",
      "id": "2iKJV1XbGfE7DhzCxS6gSuP8mK2",
      "title": "A book",
      "updatedAt": "2024-06-24T12:49:13.964459Z"
    }
]

### DIFF
--- a/integration/testdata/embedded_data/main.test.ts
+++ b/integration/testdata/embedded_data/main.test.ts
@@ -14,7 +14,6 @@ test("get action with embedded data", async () => {
 
   const fetchedPost = await actions.getPost({ id: post.id });
   expect(fetchedPost!.id).toEqual(post.id);
-  expect(fetchedPost!.categoryId).toBeUndefined();
   expect(fetchedPost!.category!.title).toEqual("Test");
 });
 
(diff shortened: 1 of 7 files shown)


davenewza (commented):
Sorry if this is misguided question, but could this not be solved in makeproto instead?  And then the API implementations should not need to be reworked?

RutZap:
> Sorry if this is misguided question, but could this not be solved in makeproto instead? And then the API implementations should not need to be reworked?

Since the embeds are defined at action level, if we're removing the foreign key from the model at proto level then we're affecting actions that don't have embeds at all.

davenewza (approved):
LGTM!

PR #1539, opened 2024-07-30
fix: handle null InlineFiles updates

RutZap:
Fix bug that was preventing updates on InlineFiles to be set as null.


davenewza (approved):
LGTM!

PR #1545, opened 2024-08-07
feat: inline files read

RutZap:
Support for InlineFile.read() from S3

### DIFF
--- a/node/codegen.go
+++ b/node/codegen.go
@@ -173,7 +173,7 @@ func writeBuiltInTypes(w *codegen.Writer) {
 	w.Writeln(`static fromObject(obj: any): InlineFile;`)
 	w.Writeln(`static fromDataURL(url: string): InlineFile;`)
 	w.Writeln(`read(): Buffer;`)
-	w.Writeln(`store(): Promise<any>;`)
+	w.Writeln(`store(expires?: Date): Promise<any>;`)
 	w.Writeln(`filename: string;`)
 	w.Writeln(`contentType: string;`)
 	w.Writeln(`size: number;`)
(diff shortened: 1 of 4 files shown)


node/codegen.go
```diff
@@ -169,11 +169,11 @@ func writeBuiltInTypes(w *codegen.Writer) {
 	w.Writeln(`export type SortDirection = "asc" | "desc" | "ASC" | "DESC"`)
 	w.Writeln(`export declare class InlineFile {`)
 	w.Indent()
-	w.Writeln(`constructor(key: any, filename: any, contentType: any, size: any, url: any);`)
+	w.Writeln(`constructor(key: any, filename: any, contentType: any, size: any, url: any, key: any);`)
 	w.Writeln(`static fromObject(obj: any): InlineFile;`)
 	w.Writeln(`static fromDataURL(url: string): InlineFile;`)
-	w.Writeln(`read(): Buffer;`)
-	w.Writeln(`store(): Promise<any>;`)
+	w.Writeln(`read(): Promise<any>;`)
```
davenewza, line 175: Can we type this?  Perhaps `byte[]`?

node/codegen.go
```diff
@@ -169,11 +169,11 @@ func writeBuiltInTypes(w *codegen.Writer) {
 	w.Writeln(`export type SortDirection = "asc" | "desc" | "ASC" | "DESC"`)
 	w.Writeln(`export declare class InlineFile {`)
 	w.Indent()
-	w.Writeln(`constructor(key: any, filename: any, contentType: any, size: any, url: any);`)
+	w.Writeln(`constructor(key: any, filename: any, contentType: any, size: any, url: any, key: any);`)
```
davenewza, line 172: Can we use `string`, `number`, instead of `any`?

davenewza (approved):
This will call for a solid backend test in `deploy_test.go` which reads and writes files in the functions runtime

PR #1550, opened 2024-08-08
feat: functions-runtime errors

tomfrew:
* Export an `errors` object from the functions SDK with error classes which can be used to throw specific error types which when get transforms into the correct status codes in the response by the runtime
* Uses `ERR_UNKNOWN` instead of `ERR_INTERNAL` for general errors in the function to separate core runtime errors from user code errors

```typescript
import { errors } from "@teamkeel/sdk";

export default CustomFunction(async (ctx, inputs) => {
    throw new errors.NotFound("custom message"); // Returns a 404 with the standard runtime error strcuture
}
```

### DIFF
--- a/functions/functions.go
+++ b/functions/functions.go
@@ -36,6 +36,7 @@ const (
 	NotNullConstraintError    FunctionErrorCode = -32006
 	UniqueConstraintError     FunctionErrorCode = -32007
 	PermissionError           FunctionErrorCode = -32008
+	BadRequestError           FunctionErrorCode = -32009
 )
 
 type FunctionType string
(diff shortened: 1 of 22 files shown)


functions/functions.go
```diff
@@ -312,13 +313,27 @@ func toRuntimeError(errorResponse *FunctionsRuntimeError) error {
 	case UniqueConstraintError:
 		return common.NewUniquenessError(strings.Split(data["column"].(string), ", "))
 	case RecordNotFoundError:
+		if errorResponse.Message != "" {
+			return common.RuntimeError{
+				Code:    common.ErrRecordNotFound,
+				Message: errorResponse.Message,
+			}
+		}
 		return common.NewNotFoundError()
```
davenewza, line 322: Could have `common.NewNotFoundError()` take in a message string argument

davenewza (approved):

functions/functions.go
RutZap: I was thinking about the same thing:

```
func NewNotFoundError(message string) RuntimeError {
	if message == "" {
		message = "record not found"
	}

	return RuntimeError{
		Code:    ErrRecordNotFound,
		Message: message,
	}
}
```

PR #1551, opened 2024-08-09
fix: reuse inlinefile transform function in jobs

davenewza:
Reusing the `InlineFile` transform function in the job handler.

### DIFF
--- a/runtime/actions/parsing.go
+++ b/runtime/actions/parsing.go
@@ -183,23 +183,15 @@ var toDate = func(value any) (types.Date, error) {
 //
 // e.g. for InlineFile inputs, whcih are given as a dataURL string, they need to be transformed into an object
 // including the typename
-func TransformCustomFunctionsInputTypes(schema *proto.Schema, action *proto.Action, input any) (any, error) {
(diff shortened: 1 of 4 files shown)


runtime/actions/parsing.go
```diff
@@ -183,29 +183,21 @@ var toDate = func(value any) (types.Date, error) {
...
-	if !isMap {
-		return input, nil
-	}
-
-	msg := proto.FindMessage(schema.Messages, action.InputMessageName)
-	if msg == nil {
-		return input, nil
-	}
+func TransformCustomFunctionsInputTypes(schema *proto.Schema, messageName string, input map[string]any) (map[string]any, error) {
+	message := proto.FindMessage(schema.Messages, messageName)
```
RutZap, line 187:
```suggestion
	message := proto.FindMessage(schema.Messages, messageName)
	if message == nil {
		return input, nil
	}
```

In the case of a custom action with `Any`, the message won't exist and the following bit of code will throw a panic (`message.HasFiles()`)

davenewza: Nice

RutZap (approved):

PR #1552, opened 2024-08-09
feat: store InlineFiles automatically when passed through to modelAPI's create

RutZap:
When creating models with the model api, InlineFiles will be automatically stored before inserting in the db.

```
export default CreateFromImage(async (ctx, inputs) => {
  const post = await models.post.create({
    image: inputs.image,
    description: inputs.description,
  })
});


davenewza (approved):
Awwwesome!

PR #1557, opened 2024-08-13
feat: modelAPI find and findMany to return instances of InlineFile

RutZap:
When using the modelAPI's `find()` and `findMany()` functions to retrieve models, any InlineFIles will now be returned as instances of the `InlineFile` class rather than plain objects. This allows custom functions like:

```
const post = await models.post.findOne({ id: inputs.id })
const contents = await post.image.read();

return {
         imageContents: contents.toString('base64'),
}
```


packages/functions-runtime/src/ModelAPI.js
```diff
@@ -327,6 +327,29 @@ async function create(conn, tableName, tableConfigs, values) {
...
+// Iterate through the given object's keys and if any of the values are a rich data type, instantiate their respective class
+function transformRichDataTypes(data) {
+  const keys = data ? Object.keys(data) : [];
+  const row = {};
+
+  for (const key of keys) {
+    const value = data[key];
+    if (isPlainObject(value)) {
+      // if we've got an InlineFile...
+      if (value.key || value.size || value.filename || value.contentType) {
```
davenewza, line 339: Shouldn't this be `if (value.key && value.size && value.filename && value.contentType) {`?

davenewza (approved):
Yaaa!  Can we get an integration test? 😆

PR #1563, opened 2024-08-22
chore: refactor proto query

RutZap:
Housekeeping update of the proto package:

Duplicated most functions as methods on their respective structures; e.g. `func (f *Field) IsHasMany() bool` instead of ` func IsHasMany(field *Field)` and deprecated original functions. Migrate all usage of newly deprecated function to the new definitions. 

Split the methods across multiple smaller and easier to follow files relevant to the receiver of the method (e.g. `model.go`, `field.go` etc. )


davenewza (approved):
Very nice.  Love the changes

PR #1565, opened 2024-08-23
chore: re-enable staticcheck linter. update deprecated method

RutZap:
Re-enabling staticcheck. Update trace check for noop to use new package


davenewza (approved):
LGTM

PR #1571, opened 2024-08-28
feat: add tools configuration generation

RutZap:
Adding a new RPC which generates and returns a configuration for tools used to interact with the current project.


node/codegen.go
```diff
@@ -1614,7 +1614,7 @@ func writeTestingTypes(w *codegen.Writer, schema *proto.Schema) {
 		w.Writeln("withIdentity(identity: sdk.Identity): JobExecutor;")
 		w.Writeln("withAuthToken(token: string): JobExecutor;")
 		for _, job := range schema.Jobs {
-			msg := proto.FindMessage(schema.Messages, job.InputMessageName)
+			msg := schema.FindMessage(job.InputMessageName)
```
davenewza, line 1617: :)

tools/generator.go
```diff
@@ -0,0 +1,660 @@
...
+type Tool struct {
+	Config         *rpc.ActionConfig
+	Model          *proto.Model
+	Action         *proto.Action
+	SortableFields []string
+}
+
+type Generator struct {
+	Schema *proto.Schema
+	Tools  map[string]*Tool
```
davenewza, line 23: Is it necessary to have a pointer to `Tool` in the map?

tools/generator.go
```diff
@@ -0,0 +1,660 @@
...
+}
+
+const fieldNameID = "id"
+
+var ErrInvalidSchema = errors.New("invalid schema")
+
+// NewGenerator creates a new tooll config generator for the given schema
```
davenewza, line 30:
```suggestion
// NewGenerator creates a new tool config generator for the given schema
```

tools/generator.go
```diff
@@ -0,0 +1,660 @@
...
+	for id, tool := range g.Tools {
+		if tool.Model.Name == modelName && tool.Action.IsGet() && tool.hasOnlyIDInput() {
+			return id
+		}
+	}
+
+	return ""
+}
+
+// findByIDTools searches for the tools that operate on the given model and take in an ID as an input; Returns a map of
```
davenewza, line 512:
```suggestion
// findAllByIDTools searches for the tools that operate on the given model and take in an ID as an input; Returns a map of
```

davenewza (approved):
Looks great

tools/generator.go
RutZap: So I think if we use pointers, it's slightly more performant when our structs are larger. Also, if we don't use pointers we can't do something like this: `g.Tools[input.GetEntryAction.ToolId].getIDInputFieldPath(),` not any direct assignments like `g.Tools[id].X = X`

RutZap: To be fair, your comment made me look deeper into it, and this [article](https://www.willem.dev/articles/slice-of-pointers-structs/) explains quite nicely the differences, especially:

> When you iterate over a slice of values, the iteration variables will be copies of those values. Any modifications you make to the iteration variables won’t be reflected outside of the loop.

> With a slice of pointers, the iteration variables will be copies of the pointers. These copies will still point to the same values. Changes to these values will be visible outside of the loop.

So yeah, I think having a map of pointers makes some parts of the code much easier to work with (i.e. decorating the tools directly by referencing the map pointer e.g. `g.Tools[id].Config.X = X`)

PR #1572, opened 2024-08-28
fix: storedfile, modelapi typing, additional files support

davenewza:
- New `StoredFile` type distinct from `InlineFile`
 - Common mime types as string type options
 - Models to use `StoredFile` type while create and update values to support `StoredFile | InlineFile` 
 - ModelAPI `update` implicit storing support and addition file support in query builder
 - Kysely configured to use `FileDbRecord`
 - Added file storage for `keel test`
 - More custom function and hooks tests


node/codegen.go
```diff
@@ -165,24 +166,6 @@ func generateSdkPackage(schema *proto.Schema, cfg *config.ProjectConfig) codegen
 	}
 }
 
-// writeBuiltInTypes will write the types for Built In types such as InlineFile, SortDirection, etc..
-func writeBuiltInTypes(w *codegen.Writer) {
-	w.Writeln(`export type SortDirection = "asc" | "desc" | "ASC" | "DESC"`)
```
RutZap, line 170: I think we need to add the `SortDirection` back in `generateSdkPackage()`

RutZap: Ah scratch that. I just saw we've added it to the template

davenewza: @RutZap Ive added it to the `functions-runtime` types since it doesnt need to be generated

RutZap (approved):

PR #1581, opened 2024-09-04
fix: presigned url in functions

davenewza:
A presigned URL can now be retrieved from an instance of the `File` class using `getPresignedUrl()`.  Just like with the Go runtime, the URL expires after 24 hours.

### DIFF
--- a/packages/functions-runtime/src/File.js
+++ b/packages/functions-runtime/src/File.js
@@ -170,7 +170,9 @@ class File extends InlineFile {
         Key: "files/" + this.key(),
       });
 
-      return new URL(getSignedUrl(s3Client, command, { expiresIn: 60 * 24 }));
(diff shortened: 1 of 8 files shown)


packages/functions-runtime/src/File.js
```diff
@@ -157,6 +158,30 @@ class File extends InlineFile {
...
   }
 
+  async getPresignedUrl() {
+    if (isS3Storage()) {
+      const s3Client = new S3Client({
+        credentials: fromEnv(),
+        region: process.env.KEEL_REGION,
+      });
+
+      const command = new PutObjectCommand({
```
RutZap, line 168:
```suggestion
      const command = new GetObjectCommand({
```

RutZap (approved):

PR #1594, opened 2024-09-11
fix: functions database connection limit and retries

davenewza:
Limiting the number of connections in the functions runtime to 1.  Furthermore, configuring 3 retries with an interval of 1s.


RutZap (approved):

packages/functions-runtime/src/database.js
```diff
@@ -185,6 +185,9 @@ function getDialect() {
 
       const pool = new InstrumentedNeonServerlessPool({
         connectionString: mustEnv("KEEL_DB_CONN"),
+        max: 1, // Limit to 1 connection per Lambda instance
+        retryInterval: 1000, // Retry every second
+        maxRetries: 3, // Maximum number of retries
```
davenewza, line 190:
```suggestion
```

PR #1596, opened 2024-09-11
fix: add model name to ID message fields when used as a _list where i…

RutZap:
…nput_ message


davenewza (approved):

davenewza:
I think we might have more cases of this

PR #1600, opened 2024-09-13
fix: updates to left join on relationships

davenewza:
When performing an `update` which requires checking conditions on related tables, we need to ensure that we are left joining on those tables.  The reason for this is because if there is no relationship present, then we don't want the row to be omitted.

However, we have been using `FROM` in our update statements.  When relationships are involved, will are now using sub-queries.  This is cleared to understand and fixes the issue.  

Except from Postgres docs: 

> When a FROM clause is present, what essentially happens is that the target table is joined to the tables mentioned in the from_item list, and each output row of the join represents an update operation for the target table. When using FROM you should ensure that the join produces at most one output row for each row to be modified. In other words, a target row shouldn’t join to more than one row from the other table(s). If it does, then only one of the join rows will be used to update the target row, but which one will be used is not readily predictable.
> Because of this indeterminacy, referencing other tables only within sub-selects is safer, though often harder to read and slower than using a join.

### DIFF
--- a/runtime/actions/query.go
+++ b/runtime/actions/query.go
@@ -969,12 +969,7 @@ func (query *QueryBuilder) UpdateStatement(ctx context.Context) *Statement {
 	//var from string
 	if len(query.joins) > 0 {
 		for _, j := range query.joins {
-			//if i == 0 {
-			//	from = fmt.Sprintf("FROM %s AS %s", j.table, j.alias)
-			//	queryFilters = append([]string{j.condition, "AND"}, queryFilters...)
-			//} else {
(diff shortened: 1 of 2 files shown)


runtime/actions/query.go
```diff
@@ -966,15 +966,10 @@ func (query *QueryBuilder) UpdateStatement(ctx context.Context) *Statement {
 
 	args = append(args, query.args...)
 
-	var from string
+	//var from string
```
RutZap, line 969:
```suggestion
```

RutZap (approved):

PR #1615, opened 2024-09-27
feat: function configuration for auto transactions

tomfrew:
We currently wrap all mutation functions in a transaction which gets rolled back if the function throws an error. This is a great default but there are times where you do want to write to the db and also return an error. This PR adds optional configuration to our generated function wrappers which would enable users to control the behaviour of the functions runtime. And then uses this config to control the automatic transaction behaviour

e.g.
```ts
MyFunction.config = {
  autoTransaction: false,
};

export default MyFunction(async (ctx, inputs) => {
  // .. do things
});

``` 

or for hooks

```ts
const hooks: MyFunctionHooks = {
  beforeWrite: async (ctx, inputs, values, record) => {
   // ..do things
  },
};

MyFunction.config = {
  autoTransaction: false,
};

export default MyFunction(hooks);

```

### DIFF
--- a/packages/functions-runtime/src/index.d.ts
+++ b/packages/functions-runtime/src/index.d.ts
@@ -227,3 +227,15 @@ export type Errors = {
    */
   Unknown: typeof UnknownError;
 };
+
+export type FunctionConfig = {
+  /**
+   * All DB calls within the function will be executed within a transaction.
+   * The transaction is rolled back if the function throws an error.
+   */
+  dbTransaction?: boolean;
+};
+
+export type FuncWithConfig<T> = T & {
+  config: FunctionConfig;
+};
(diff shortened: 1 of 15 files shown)


packages/functions-runtime/src/index.d.ts
```diff
@@ -227,3 +227,15 @@ export type Errors = {
    */
   Unknown: typeof UnknownError;
 };
+
+export type FunctionConfig = {
+  /**
+   * All DB calls within the function will be executed within a transaction.
+   * The transaction is rolled back if the function throws an error.
+   */
+  autoTransaction?: boolean;
```
RutZap, line 236:
```suggestion
  dbTransaction?: boolean;
```

packages/functions-runtime/src/tryExecuteFunction.js
```diff
@@ -11,11 +11,24 @@ const { PROTO_ACTION_TYPES } = require("./consts");
...
+      case PROTO_ACTION_TYPES.GET:
+      case PROTO_ACTION_TYPES.LIST:
+      case PROTO_ACTION_TYPES.READ:
+        requiresTransaction = false;
+        break;
+    }
+
+    if (functionConfig?.autoTransaction !== undefined) {
+      requiresTransaction = functionConfig.autoTransaction;
+    }
```
RutZap, line 29:
```suggestion
    if (functionConfig?.dbTransaction !== undefined) {
      requiresTransaction = functionConfig.dbTransaction;
    }
```

RutZap (approved):

RutZap (approved):

PR #1624, opened 2024-10-07
feat: sync watcher exclusions with tsconfig.json

xakdog:
## Goal
Simplify configuration for mono repos with Keel and front-end projects

## Problem
1. Keel requires `schema.keel` at the project root in mono repos
2. To prevent Keel from treating front-end JS as functions, we exclude directories in `tsconfig.json`
3. However, the watcher still triggers on these excluded directories

## Solution
Make the watcher respect `exclude` paths from `tsconfig.json`

## Note
This change only affects watcher behavior, not compilation or deployment processes

### DIFF
--- a/cmd/program/commands.go
+++ b/cmd/program/commands.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/rsa"
 	"crypto/x509"
+	"encoding/json"
 	"encoding/pem"
 	"errors"
 	"fmt"
@@ -19,6 +20,7 @@ import (
 	_ "embed"
 
 	"github.com/Masterminds/semver/v3"
+	"github.com/bmatcuk/doublestar/v4"
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/radovskyb/watcher"
 	"github.com/teamkeel/keel/cmd/cliconfig"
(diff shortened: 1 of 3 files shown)


RutZap:
@xakdog Thank you so much for contributing! We really like the approach and picking up the `exclude` paths from `tsconfig` is great. 

I've had a quick play with this and it seems that there are some issues with the pattern matching; `keel` allows us to pass a base path to the project as a cli arg; e.g. `keel run -d ./../../path/to/project`. When using it like so, it seems that the watcher fails to ignore the `.build` folder which then it causes it to get stuck in a continuous rebuild loop. I'm not 100% sure what the fix is at the moment; as soon as we fix this issue we'll look at merging this through.

RutZap (changes requested):
@xakdog Thank you so much for contributing! We really like the approach and picking up the `exclude` paths from `tsconfig` is great.

I've had a quick play with this and it seems that there are some issues with the pattern matching; keel allows us to pass a base path to the project as a cli arg; e.g. `keel run -d ./../../path/to/project`. When using it like so, it seems that the watcher fails to ignore the .build folder which then it causes it to get stuck in a continuous rebuild loop. I'm not 100% sure what the fix is at the moment; as soon as we fix this issue we'll look at merging this through.

jonbretman:
Hey @xakdog - we believe we have addressed this issue now as part of some other changes so I'm going to close this. Thanks again for the PR and do let us know if it's still an issue.

PR #1626, opened 2024-10-14
chore: refactor keelconfig validation to use json schema

jonbretman:
Hello 👋 

In looking at adding some stuff to keelconfig as part of getting a `keel deploy` command I was looking at how we validate the config and felt that JSON schema would be a better approach. I then discovered that `keel validate` wasn't actually validating the config at all.

* Move (most) validation to JSON schema
* Fix `keel validate` to validate all keelconfig files

### Config errors
Screenshot 2024-10-14 at 10 46 29

### As JSON
Screenshot 2024-10-14 at 10 46 58

### Later...
We're not currently doing anything with the config errors in the VSCode extension that I can see, so that needs doing. I know how to get line/pos info for each error but not done in the PR. I'll add it later and then we can put the redd squigglies in the right place 😄


RutZap (approved):
💪

PR #1639, opened 2024-10-21
chore: tools - add display layout proto definitions

RutZap:
Introducing a new config option for tools: `DisplayLayoutConfig`. This allows configuring the way generated tools are displayed, with the aim of supporting multiple types of layouts: inbox, data-grid, kanban, etc


tools/proto/tools.proto
```diff
@@ -279,3 +285,34 @@ message DataMapping {
...
+message DisplayLayoutConfig {
+	enum Type {
+		UNKNOWN = 0;
+		INBOX = 1;
+	}
+
+	// The type of the view: e.g. Inbox, List, Calendar, Table, Data Grid
+	Type type = 1;
+
+	optional InboxViewConfig inbox_config = 2;
```
davenewza, line 298: Ah, I see.  So this is where we'll be adding the different view configs 👍

tools/proto/tools.proto
```diff
@@ -279,3 +285,34 @@ message DataMapping {
...
+	// The secondary content of the row
+	optional StringTemplate description = 3; 
+	// Content to be displayed in the right hand side of the row
+	optional StringTemplate context_field = 4;
+	
+	// If an avatar/image is to be displayed with every row
+	bool avatar_enabled = 5;
+	// URL for the image.
+	optional StringTemplate avatar_url = 6;
+	optional StringTemplate avatar_fallback = 7;
```
davenewza, line 317: What is avatar_fallback again?

davenewza (approved):

tools/proto/tools.proto
RutZap: So in the case of the avatar_url being a string template like `{{$.avatar}}`.. and `avatar` being a optional field  type `File`, if there would be a user without an image, the fallback could be used to still render an avatar; i.e. a placeholder image generated with the user's initials; i.e. `https://ui-avatars.com/api/?name={{$.userName}}`

davenewza: Ah I see.  @RutZap I think this as a comment would be super useful

PR #1640, opened 2024-10-22
feat: show annotated source for keelconfig errors in keel validate command

jonbretman:
By switching out `"gopkg.in/yaml.v3"` for `"github.com/goccy/go-yaml"` we can have pretty annotated errors and line/column info for Keel config errors.

#### Errors in different files
Example of errors in different files

#### With schema errors
Example of both schema + config errors
 in schema + config

#### JSON output
We can use this in the VSCode extension to show squiggles in the YAML
Example of JSON output


davenewza (approved):
LGTM.  Super improvements

PR #1666, opened 2024-11-05
WIP: expression validation and evaluation with CEL

davenewza:
Expression parsing using CEL
===

📖 See Computed Fields product doc here:  https://www.notion.so/keelhq/Computed-fields-fa1f91f188134ece9b2072e9b22e0c1e

Why?
---

Computed fields will introduce a lot more complexity to Keel expressions.  We'll need to support:
 - **Arithmetic operations,** e.g. `item.price - item.cost`
 - **Functions with arguments**, e.g. `ABS(value)` or `DURATION(start, end)` or `YEAR(date)`
 - **Aggregate functions which operate on 1:M data**, e.g. `SUM(items.price)`
 - **Expression nesting with proper operator/function precedence**, e.g. `SUM(account.transactions.value) + account.overdraft > 0 or account.isLocked`

Our previous expression AST and parsing implementation would need to be significantly extended to support these new features and could would become quite a substantial amount of code to write and maintain, especially if we want to have really good type checking and validations.

The [cel-go package](https://github.com/google/cel-go) parses, validates and evaluates [CEL](https://github.com/google/cel-spec) expressions and also supports everything raised above.  CEL is _almost_ identical to Keel expressions, and we've proposed adopting it so not to spend precious energy on extending our codebase to do almost exactly what cel-go would do for us.

Technical overview
---

The main objective of this PR is to reach **feature parity** and not to introduce anything new yet to our expressions language.

These points mostly cover what has changed in this PR:
 - Our AST doesn't describe expressions by their distinct parts anymore.
 - Each attribute has its own cel-go _configuration_ used for parsing.
 - Expression validation is now done by cel-go.  We do not need to interrogate the AST anymore.
 - Validation errors from cel-go are intercepted and made more friendly.  Fortunately we still support discrete token positions with these validation messages.
 - Expression evaluation (for example, to generate SQL) is done using _visitors_.
 - SQL gen tests for `@where`, `@set` and `@permission` all remain mostly unchanged.

Primary packages
---

I think it's useful to briefly explain the main packages involved.

`expressions`:  The abstraction over CEL which can validate expressions with `Validate(*parser.Expression)`

`expressions/options`:  A library of options which can be used to configure the validator. That's important because each attribute would configure their expressions with more or less context and features.  For example, a `@where` expression would be configured with its action input variables and with `ctx` whereas a `@default` attribute would not.

`expressions/visitor`:  An expression visitor runner which can be used to interrogate expressions if need be.  For example, this can be used to generate SQL or read idents (see next).

`expressions/resolve`:  Various ways to resolve expressions down to values, idents, ident arrays, etc.  Sometimes we actually only care about _constants_ and not evaluation, for example, `@unique([sku, supplierCode])`. 

`parser/attributes`:  Parser configuration for each attribute type.  Our validation code only needs to really call on these to validate their expressions. E.g. `issues, err := attributes.ValidateWhereExpression(asts, action, expression)`

What's still to do?
---

 

 - [x] Validation messages need to be improved as they are less useful than what they used to be.  
 - [ ] Validation message hints need to be reintroduced.
 - [x] The functions runtime permission generation needs to be supported.  This should be simple enough to do by implementing a new visitor.
 - [x] DateTime and Date comparison support needs to be added to the cel-go configuration options.
 - [ ] Optional inputs and null fields need stricter validation checking still.
 - [x] `not in` is currently not supported. 
 - [x] Migrating to using `&&` and `||` - breaking change - more to follow.
 - [x] There is a bug with token positions.
 - [x] Fix the rest of the schema validation tests 

Removed validations
---
```
model Post {
    @permission(
        //expect-error:34:39:E032:Cannot have mixed types in an array literal
        expression: "bob" in [1, "123"],
        actions: [get]
    )
}
```

```
model Person {
    fields {
        name Text
        employer Company?
    }

    actions {
        create createPerson() with (employer.name) {
            //expect-error:37:56:E060:name cannot be null
            @permission(expression: person.name != null)
        }
        list listPersons() {
            //expect-error:20:39:E060:name cannot be null
            @where(person.name == null)
            //expect-error:20:39:E060:name cannot be null
            @where(null == person.name)
            //expect-error:20:48:E060:name cannot be null
            @where(person.employer.name == null)
            //expect-error:37:56:E060:name cannot be null
            @permission(expression: person.name != null)
        }
    }
}
```

```
model Person {
    actions {
        get getPerson(id)
    }

    @permission(
        expression: true,
        //expect-error:18:29:AttributeExpressionError:expression expected to resolve to type ActionType[] but it is but it is an array of different types
        actions: [get, true]
    )
}

```


schema/parser/expressions.go
```diff
@@ -157,110 +80,45 @@ func ParseExpression(source string) (*Expression, error) {
...
+var ErrInvalidAssignmentExpression = errors.New("expression is not a valid assignment")
 
-			if andExpr.Condition.RHS != nil {
-				result += andExpr.Condition.RHS.ToString()
-			}
-		}
+// ToAssignmentExpression splits an assignment expression into two seperate expressions.
+// E.g. the expression `post.age = 1 + 1` will become `post.age` and `1 + 1`
+func (expr *Expression) ToAssignmentExpression() (*Expression, *Expression, error) {
+	parts := strings.Split(expr.String(), "=")
```
jonbretman, line 88: An edge case but wouldn't this break with an expression like `model.field = "=== HELLO ==="`

schema/testdata/errors/attribute_expression_unresolvable_rhs.keel
```diff
@@ -2,20 +2,20 @@ model Post {
...
+        someId ID
     }
 
     actions {
         update updatePost(id) with (title) {
             //expect-error:18:28:ActionInputError:title is already being used as an input so cannot also be used in an expression
             @set(post.title = title)
         }
+        update updatePost2(id) with (title) {
+            //expect-error:18:44:AttributeExpressionError:undeclared reference to 'post' (in container '')
```
jonbretman, line 15: You're not wrong about the error messages being less helpful. The error here is that it's using `post` on the RHS which isn't allowed right? How can we make these errors better? `in container ''` is pretty cryptic

jonbretman (commented):
I do worry we're going to make the validation errors a lot less friendly, which will hurt our less technical users more maybe if they mostly work in the schema.

Do you have a plan for how we can make them better?

schema/parser/expressions.go
davenewza: @jonbretman Ah yes, what I've done is quite crude.  I should rather find the assignment operator by iterating through the tokens.  Will do that 👍

schema/testdata/errors/attribute_expression_unresolvable_rhs.keel
davenewza: Ah yes, this is actually fixed - Im still in the progress of updating all these tests.  It's actually erroring with `unknown identifier 'post'`, but even that can be improved further

runtime/actions/writeinputs.go
```diff
@@ -3,58 +3,49 @@ package actions
...
-		if !lhsResolver.IsModelDbColumn() {
-			return errors.New("lhs operand of assignment expression must be a model field")
-		}
-
-		fragments, err := lhsResolver.NormalisedFragments()
+		fragments, err = normalisedFragments(scope.Schema, fragments)
 		if err != nil {
 			return err
 		}
+		// fragments, err := lhsResolver.NormalisedFragments()
```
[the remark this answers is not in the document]
davenewza: Ta

permissions/permissions.go
```diff
@@ -93,56 +93,56 @@ func ToSQL(s *proto.Schema, m *proto.Model, action *proto.Action) (sql string, v
...
-			}
-
-			if opClose != "" {
-				stmt.expression += opClose
-			}
-		}
-	}
-
-	stmt.expression += ")"
+	// stmt.expression += "("
```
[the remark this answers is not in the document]
davenewza: Yeah it is.  I need to fit this code into a cel visitor which should be relatively straight forward.

expressions/parser_test.go
```diff
@@ -0,0 +1,843 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/teamkeel/keel/expressions"
+	"github.com/teamkeel/keel/expressions/options"
+	"github.com/teamkeel/keel/schema/parser"
+	"github.com/teamkeel/keel/schema/reader"
+	"github.com/teamkeel/keel/schema/validation/errorhandling"
+)
+
+func TestParser_Variable(t *testing.T) {
```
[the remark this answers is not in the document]
davenewza: That's a great point.  The only reason why I didn't do this is because I've been going about this PR with a bit of a test-driven approach and just happened to not use fixtures, but agreed that this needs to change. Will add it to the list

davenewza:
> I do worry we're going to make the validation errors a lot less friendly, which will hurt our less technical users more maybe if they mostly work in the schema.
> 
> Do you have a plan for how we can make them better?

I agree that this will be problem if not addressed.  My plan is to expand on converting cel-go validation errors to customer-friendly ones.  This is working pretty well so far [here](https://github.com/teamkeel/keel/pull/1666/files#diff-f4dfd3d76ba42915bee7ff8ce7e9bed4fd096feb8c688684743a0c604ec51b11) but as mentioned needs to be expanded on.  I also need to reintroduce hints for each attribute type.

expressions/typing/provider.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/teamkeel/keel/schema/parser"
+	"github.com/teamkeel/keel/schema/query"
+	expr "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+)
+
+// TypeProvider supplies the CEL context with the relevant Keel types and identifiers
+type TypeProvider struct {
+	Schema  []*parser.AST
+	Model   string
+	Objects map[string]map[string]*types.Type
```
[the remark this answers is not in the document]
davenewza: Will do - thanks

schema/parser/expressions.go
davenewza: Improved this now

cron/cron.go
```diff
@@ -599,9 +599,7 @@ func toTokens(src string) *Tokens {
 				char == ',') // [4]
 
 		if isEndOfToken {
-			// +2 here because we want the end position to point to the character
-			// after the final character of this token
-			t.End = i + 2
```
davenewza, line 604: I can't seem to figure out why I needed to remove this to the positioning correct

davenewza:
See https://github.com/teamkeel/keel/pull/1685

PR #1667, opened 2024-11-05
fix: add embed attribute to code completions

RutZap:
`@embed` was previously not included in code completions for attributes within a action block. This now fixes the issue.


davenewza (approved):
Nice 👍

PR #1668, opened 2024-11-12
feat: allow the setting of db connection and ca file in functions database

jonbretman:
This is a bit of work that is needed to support `keel deploy`, specifically RDS.

The main thing here is making it possible to set the cert, which is needed for RDS to enforce SSL.

While I was here I've made it possible to set the connection string without it needing to be an env var, which is good for not exposing db credentials. The env var still works for backwards compatibility.


davenewza (approved):
Looks great

jonbretman:
I'm going to have a go at making the passing of the connection string a little better - hold off on merging for now.

jonbretman:
@davenewza ok much better now - think this is good to go. Once this is shipped can update the runtime to pass the db connection via a secret and then eventually remove the env var.

PR #1674, opened 2024-11-26
feat: client - automatic time-zone header

tomfrew:
* In the generated client, try to get the user's timezone. If available add it as the time-zone header to support relative dates.
* Can be overridden by the config


jonbretman:
> Can be overridden by the config

Can it? Where?

tomfrew:
By setting `config.headers`

RutZap (approved):

PR #1675, opened 2024-11-27
fix: add validation rule for duplicate API models

RutZap:
When defining APIs and including models, we now validate that a model hasn't been included multiple times e.g:

```keel
model Post {
}
...

api Api {
    models {
        Post
        Post
    }
}
```

will now generate a validation error:

```
❌ The following errors were found in your schema files:

  | ../projects/keel-project/schema.keel
  38|     models {
  39|         Post
  40|         Post
    |         ──┬─
    |           ╰─ The model 'Post' has already been included in the 'Api' API (DuplicateDefinitionError)
    |              Remove one of the duplicates
  41|     }
  42| }
  43| 
```


davenewza (changes requested):
As discussed, would be best to use the visitor validation pattern as we're slowly moving away from this way of doing it.  Likewise with the error template messages

PR #1681, opened 2024-12-04
fix: generate the client api with proxy that calls into rawrequest to save bundle size

jmeistrich:
This changes the generated APIClient to not need to include all of the actions on the JavaScript object. The generated actions were all doing exactly the same thing, calling `rawRequest` with its own function name and a single parameter. So this makes the APIClient a Proxy which for any child name returns a function that calls `rawRequest` with that name. And it still generates all the types so the Proxy has the exact same type shape as the APIClient did before.

This reduces bundle size significantly, especially with large API surfaces. Rather than scaling with the number of actions it is now just a constant sized small Proxy object. For example the APIClient in Bravely has 213 actions, and the generated APIClient (after types compiled out) was 41kb (4kb gzipped). Now it's 343 bytes (195 bytes gzipped).

### DIFF
--- a/node/client.go
+++ b/node/client.go
@@ -54,7 +54,6 @@ func generateClientSdkFile(schema *proto.Schema, api *proto.Api) codegen.Generat
 	client.Writeln("// API")
 	client.Writeln("")
 
-	writeClientApiInterface(client, schema, api)
 	writeClientApiClass(client, schema, api)
 
 	return []*codegen.GeneratedFile{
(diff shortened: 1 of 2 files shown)


davenewza:
@jmeistrich thanks for this! Would you mind providing a description of what this is about before we review it?

jmeistrich:
@davenewza Sure! I updated the description.

davenewza (approved):

PR #1682, opened 2024-12-06
fix: secrets in modelapi as string

davenewza:
Fields of type `Secret` were of type `any` in the ModelAPI table interface.  This has now be fixed to be `string`.  `Secret` fields are also not made available as a filter parameter.


davenewza:
This actually might not make sense.  What we probably want is some implicit way to hash the a secret value when set via the ModelAPI and then some way to compare those secret fields with a string.

PR #1684, opened 2024-12-16
fix: add board view support to tools and dialog flag on actionlinks

RutZap:
Tools Proto schema update:

Adding a new display layout option for tools: Board View. This emulates a kanban style-view and provides all the configuration options needed to support a tool of this nature:

```proto
// A kanban-like board display for list views. Each item in a list will be displayed as a card within
// swimlanes generated by a group_by_field
message BoardViewConfig {
	// A link to a get action that will be used when the item is expanded
	ActionLink get_tool = 1; 
	// The primary content of the card
	StringTemplate title = 2;
	// The secondary content of the card
	optional StringTemplate description = 3; 
	// Content to be displayed in the right hand side of the card
	optional StringTemplate context_field = 4;
	
	// If an avatar/image is to be displayed with every card
	bool avatar_enabled = 5;
	// URL for the image.
	optional StringTemplate avatar_url = 6;
	optional StringTemplate avatar_fallback = 7;

	// The response field that will be used to group items within swimlanes
	JsonPath group_by_field = 8;

	// An optional update action that will be used when an item will move between swimlanes
	optional ActionLink update_action = 9;
}
```

Alongside this, we're now introducing a optional bool flag on ActionLinks which indicate wether the linked tool should be displayed as a dialog/modal or not:

```proto
// If this link should be opened up as a dialog/modal.
optional bool as_dialog = 7;
```


davenewza (approved):
LGTM

PR #1685, opened 2024-12-19
feat: expression validation and evaluation with cel

davenewza:
Expression parsing using CEL
===

📖 See Computed Fields product doc here:  https://www.notion.so/keelhq/Computed-fields-fa1f91f188134ece9b2072e9b22e0c1e

Why?
---

Computed fields will introduce a lot more complexity to Keel expressions.  We'll need to support:
 - **Arithmetic operations,** e.g. `item.price - item.cost`
 - **Functions with arguments**, e.g. `ABS(value)` or `DURATION(start, end)` or `YEAR(date)`
 - **Aggregate functions which operate on 1:M data**, e.g. `SUM(items.price)`
 - **Expression nesting with proper operator/function precedence**, e.g. `SUM(account.transactions.value) + account.overdraft > 0 or account.isLocked`

Our previous expression AST and parsing implementation would need to be significantly extended to support these new features and could would become quite a substantial amount of code to write and maintain, especially if we want to have really good type checking and validations.

The [cel-go package](https://github.com/google/cel-go) parses, validates and evaluates [CEL](https://github.com/google/cel-spec) expressions and also supports everything raised above.  CEL is _almost_ identical to Keel expressions, and we've proposed adopting it so not to spend precious energy on extending our codebase to do almost exactly what cel-go would do for us.

Technical overview
---

The main objective of this PR is to reach **feature parity** and not to introduce anything new yet to our expressions language.

These points mostly cover what has changed in this PR:
 - Our AST doesn't describe expressions by their distinct parts anymore.
 - Each attribute has its own cel-go _configuration_ used for parsing.
 - Expression validation is now done by cel-go.  We do not need to interrogate the AST anymore.
 - Validation errors from cel-go are intercepted and made more friendly.  Fortunately we still support discrete token positions with these validation messages.
 - Expression evaluation (for example, to generate SQL) is done using _visitors_.
 - SQL gen tests for `@where`, `@set` and `@permission` all remain mostly unchanged.

Primary packages
---

I think it's useful to briefly explain the main packages involved.

`expressions`:  The abstraction over CEL which can validate expressions with `Validate(*parser.Expression)`

`expressions/options`:  A library of options which can be used to configure the validator. That's important because each attribute would configure their expressions with more or less context and features.  For example, a `@where` expression would be configured with its action input variables and with `ctx` whereas a `@default` attribute would not.

`expressions/visitor`:  An expression visitor runner which can be used to interrogate expressions if need be.  For example, this can be used to generate SQL or read idents (see next).

`expressions/resolve`:  Various ways to resolve expressions down to values, idents, ident arrays, etc.  Sometimes we actually only care about _constants_ and not evaluation, for example, `@unique([sku, supplierCode])`. 

`parser/attributes`:  Parser configuration for each attribute type.  Our validation code only needs to really call on these to validate their expressions. E.g. `issues, err := attributes.ValidateWhereExpression(asts, action, expression)`

What else have we gained in this PR?
---
 - Better type compatibility.  For example, comparisons and assignment between Markdown and Text.
 - The not `!` operator in expressions.
 - Arrays fields which are `null` are now never returned during `any` and `all` queries.

Removed validations cases
---
I've had to remove the following test cases for now.  I will work on these in subsequent PRs.

```
model Post {
    @permission(
        //expect-error:34:39:E032:Cannot have mixed types in an array literal
        expression: "bob" in [1, "123"],
        actions: [get]
    )
}
```

```
model Person {
    fields {
        name Text
        employer Company?
    }

    actions {
        create createPerson() with (employer.name) {
            //expect-error:37:56:E060:name cannot be null
            @permission(expression: person.name != null)
        }
        list listPersons() {
            //expect-error:20:39:E060:name cannot be null
            @where(person.name == null)
            //expect-error:20:39:E060:name cannot be null
            @where(null == person.name)
            //expect-error:20:48:E060:name cannot be null
            @where(person.employer.name == null)
            //expect-error:37:56:E060:name cannot be null
            @permission(expression: person.name != null)
        }
    }
}
```

```
model Person {
    actions {
        get getPerson(id)
    }

    @permission(
        expression: true,
        //expect-error:18:29:AttributeExpressionError:expression expected to resolve to type ActionType[] but it is but it is an array of different types
        actions: [get, true]
    )
}
```

Removed expression source definitions
---
Previously you could go to the definition (i.e. the model field) of in an expression ident. I've had to remove this functionality, but I will readd it once we introduce positions/node data at the ident fragment level.  This was an oversight.


schema/attributes/composite_unique.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/teamkeel/keel/expressions"
+	"github.com/teamkeel/keel/expressions/options"
+	"github.com/teamkeel/keel/schema/parser"
+	"github.com/teamkeel/keel/schema/query"
+	"github.com/teamkeel/keel/schema/validation/errorhandling"
+)
+
+func ValidateCompositeUnique(model *parser.ModelNode, expression *parser.Expression) ([]*errorhandling.ValidationError, error) {
+	opts := []expressions.Option{
+		options.WithReturnTypeAssertion("_FieldName", true),
```
RutZap, line 13: Would it not be safer to define `"_FieldName"` as a constant within the expressions package?

schema/attributes/permission.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return p.Validate(expression)
+}
+
+func ValidatePermissionActions(expression *parser.Expression) ([]*errorhandling.ValidationError, error) {
+	opts := []expressions.Option{
+		options.WithConstant(parser.ActionTypeGet, "_ActionType"),
```
RutZap, line 53: Would it not be safer to define `"_ActionType"` as a constant within the expressions package?

schema/attributes/composite_unique.go
davenewza: Yeah, would like to refine that a bit more at some point.  I have a few things in my head that I'd like to do here still

RutZap (approved):

PR #1686, opened 2024-12-20
feat: @computed attribute config and validations

davenewza:
Parsing configuration for `@computed` attribute and validations.


proto/schema.proto
```diff
@@ -78,6 +78,14 @@ message Field {
     // and then Author has posts which is of type Post, on the Post.author field this
     // value will be "posts" and on the Author.posts field this value will be "author".
     google.protobuf.StringValue inverse_field_name = 12;
+
+    // If true then this is a computed field which means its value is derived from an expression
+    // and cannot be mutated directly. It is, however, still a database column.
+    bool computed = 13;
```
jonbretman, line 84: Not sure we need both these fields. Just having `computed_expression` is probably enough and if that is not-nil then the field is computed.

schema/testdata/errors/attribute_computed.keel
```diff
@@ -0,0 +1,22 @@
+model Item {
+    fields {
+        //expect-error:23:32:AttributeArgumentError:0 argument(s) provided to @unique but expected 1
```
jonbretman, line 3:
```suggestion
        //expect-error:23:32:AttributeArgumentError:0 argument(s) provided to @computed but expected 1
```

schema/validation/computed_attribute.go
```diff
@@ -0,0 +1,92 @@
...
+					),
+				)
+			}
+
+			if len(attr.Arguments) != 1 {
+				errs.AppendError(
+					errorhandling.NewValidationErrorWithDetails(
+						errorhandling.AttributeArgumentError,
+						errorhandling.ErrorDetails{
+							Message: fmt.Sprintf("%v argument(s) provided to @unique but expected 1", len(attr.Arguments)),
```
jonbretman, line 60:
```suggestion
							Message: fmt.Sprintf("%v argument(s) provided to @computed but expected 1", len(attr.Arguments)),
```

proto/schema.proto
davenewza: Agreed

PR #1704, opened 2025-01-23
feat: computed fields and many to one relationships

davenewza:
`@computed` and M:1 relationships
===

This PR provides support for referencing fields in a M:1 relationship in a `@computed`  expressions.  For example:

```
model Item {
    fields {
        quantity Number
        product Product
        total Decimal @computed(item.quantity * item.product.price)
    }
}

model Product {
    fields {
        name Text
        price Decimal
    }
}
````

### DIFF
--- a/migrations/sql.go
+++ b/migrations/sql.go
@@ -231,11 +231,15 @@ func alterColumnStmt(modelName string, field *proto.Field, column *ColumnRow) (s
 	return strings.Join(stmts, "\n"), nil
 }
 
+func hashOfExpression(expression string) string {
+	return fmt.Sprintf("%x", sha256.Sum256([]byte(expression)))[:8]
+}
+
 // computedFieldFuncName generates the name of the a computed field's function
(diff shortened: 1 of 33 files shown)

### CODE REVIEW

File: migrations/sql.go [hunk 1]
```diff
@@ -248,6 +252,12 @@ func computedTriggerName(model *proto.Model) string {
 	return fmt.Sprintf("%s__computed_trigger", strcase.ToSnake(model.Name))
 }
 
+func computedDependencyFuncName(model *proto.Model, dependentModel *proto.Model, fragments []string) string {
```
RutZap, line 255: Just a thought here, I think by default postgres has an identifier limit of 63 bytes, I wonder if `%s__to__%s__%s__computed_dependency` may lead us to very long names that will be truncated by the db engine itself, I wonder if it's worth adding a test to assert what our limit is in a general use case scenario.

RutZap: Maybe we change `computed_dependency` to `comp_dep` or something like that to give us more space for field names

RutZap (approved):

PR #1705, opened 2025-01-23
feat: add tools sections & conditional visibility expressions to fields

RutZap:
Adding proto support for tools sections:

```proto
// A section represents a grouping of fields.
message Section {
	// A name that's referenced by response/request fields. Acts as an ID.
	string name = 1;

	// A title that will be displayed with this section
	optional StringTemplate title = 2;

	// A description for this section. Can be used to display user help text or any other copy
	optional StringTemplate description = 3;

	// Ordering of sections within a tool
	int32 display_order = 4;

	// A CEL expression used to control the visibility of the whole section.
	// e.g. if expression != true then the section will be hidden
	optional string visible_condition = 5;
}
```

Fields are added to sections by referencing the section name:

```proto
message ResponseFieldConfig {
	...

	// The section field allows the grouping of responses into sections for a leaner and better UI/UX.
	// References a Section's name.
	optional string section_name = 16;
}
```

Also adding a visibility expression to request fields and response fields:

```proto
	// A CEL expression used to control the visibility of the field.
	// e.g. if expression is not true then the field will be hidden. 
	optional string visible_condition = 17;
```


davenewza (approved):
LGTM

PR #1715, opened 2025-01-31
fix: bug with computed and one to one

davenewza:
Fixing the bug that Radu found when using a field name referencing a model which has a different name to the model.

### DIFF
--- a/runtime/actions/generate_computed.go
+++ b/runtime/actions/generate_computed.go
@@ -189,18 +189,23 @@ func (v *computedQueryGen) VisitIdent(ident *parser.ExpressionIdent) error {
 			// Join together all the tables based on the ident fragments
 			model = v.schema.FindModel(field.Type.ModelName.Value)
 			query := NewQuery(model)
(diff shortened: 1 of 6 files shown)


runtime/actions/generate_computed.go
```diff
@@ -189,18 +189,23 @@ func (v *computedQueryGen) VisitIdent(ident *parser.ExpressionIdent) error {
...
 
 			// Select the column as specified in the last ident fragment
 			fieldName := normalised[len(normalised)-1]
-			fragments := normalised[1 : len(normalised)-1]
+			fragments := subFragments[:len(subFragments)-1]
 			query.Select(ExpressionField(fragments, fieldName, false))
 
 			// Filter by this model's row's ID
-			relatedModelField := proto.FindField(v.schema.Models, v.model.Name, normalised[1])
+			//relatedModelField := proto.FindField(v.schema.Models, v.model.Name, r.Name)
```
RutZap, line 208:
```suggestion
```

PR #1719, opened 2025-02-10
feat: upgraded kysely

davenewza:
Upgraded Kysely and deprecated `orWhere` in the ModelAPI
===

Kysely updated to 0.27.5, which included some breaking changes to their API.  This unlocks Kysely's expression builder updates.

We have also deprecated the `orWhere` function on the ModelAPI.

### DIFF
--- a/integration/testdata/simple_database_test/tests.test.ts
+++ b/integration/testdata/simple_database_test/tests.test.ts
@@ -38,46 +38,6 @@ test("findMany", async () => {
   expect(results.length).toEqual(2);
 });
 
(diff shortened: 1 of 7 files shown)


jonbretman:
@davenewza what's happening with this PR? I'm assuming we don't want to drop the `orderBy` and `limit` functions from the query builder?

davenewza:
> @davenewza what's happening with this PR? I'm assuming we don't want to drop the `orderBy` and `limit` functions from the query builder?

@jonbretman They have been commented out for an age and I dont know why.  I'll try look into it today

PR #1721, opened 2025-02-11
Offset pagination

RutZap:
Introducing support for offset pagination for list actions.

This PR introduces 2 new pagination fields for list actions: `limit` and `offset`. When a `limit` is passed through, the pagination will switch from cursor to offset pagination.

The response will also include a new field within `pageInfo` : `pageNumber`. This is calculated by taking in account the given `limit` and `offset`.


davenewza (approved):
Looks good.  We just need some docs now

PR #1723, opened 2025-02-13
fix: validator improvements

RutZap:
Improvements to the keel validator:

* Explicitly check update actions to make sure there are no nested fields used as inputs as the runtime currently doesn't support updating related models; e.g.

```erlang
model Post {
  fields {
    name Text
    categ Category
  }
  actions {
    update updatePost(id) with (name, categ.name) <-- this is not allowed
  }
}

model Category {
  fields {
    name Text
    posts Post[]
  }
}
```

### DIFF
--- a/schema/validation/unique_attribute.go
+++ b/schema/validation/unique_attribute.go
@@ -162,22 +162,20 @@ func UniqueAttributeRule(asts []*parser.AST, errs *errorhandling.ValidationError
 
 			idents, err := resolve.AsIdentArray(expression)
 			if err != nil {
-				if err != nil {
-					errs.AppendError(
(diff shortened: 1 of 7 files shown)


schema/validation/unique_attribute.go
```diff
@@ -162,22 +162,20 @@ func UniqueAttributeRule(asts []*parser.AST, errs *errorhandling.ValidationError
 
 			idents, err := resolve.AsIdentArray(expression)
 			if err != nil {
-				if err != nil {
```
davenewza, line 165: Whoops!

davenewza (approved):
LGTM

PR #1727, opened 2025-02-18
perf: improving cel env setup performance

davenewza:


schema/parser/parser.go
```diff
@@ -379,5 +380,7 @@ func Parse(s *reader.SchemaFile) (*AST, error) {
 		return schema, err
 	}
 
+	schema.Raw = s.Contents
```
davenewza, line 383: Not super happy about this.  There are some other bigger performance improvements we can do though, but this should tie us over for a while

RutZap (approved):

PR #1736, opened 2025-02-23
feat: allow setting a custom endpoint for S3 client

jonbretman:
I've pulled this change out from #1678 as I've run into an issue where trying to make changes to both JS packages and then relying on those changes in integration tests doesn't always work.

The issue is that although when we install functions-runtime and testing-runtime packages into the test directories we reference the local packages, the testing-runtime package references the [published](https://github.com/teamkeel/keel/blob/main/packages/testing-runtime/package.json#L21) version of functions-runtime in it's package.json, which doesn't have the local changes.

There is probably a better way to solve this (maybe with workspaces) but for now I think the easiest thing is to merge and publish this change and then I can just update the version in #1678.

### DIFF
--- a/packages/functions-runtime/src/File.js
+++ b/packages/functions-runtime/src/File.js
@@ -9,6 +9,37 @@ const { useDatabase } = require("./database");
 const { DatabaseError } = require("./errors");
 const KSUID = require("ksuid");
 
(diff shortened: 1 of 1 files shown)


davenewza (approved):
LGTM

PR #1738, opened 2025-02-24
build: parser and proto changes to support route functions

jonbretman:
Just the basic bits for supporting defining route functions in the schema.


schema/validation/route_functions.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/samber/lo"
+	"github.com/teamkeel/keel/schema/parser"
+	"github.com/teamkeel/keel/schema/validation/errorhandling"
+)
+
+func RouteFunctions(asts []*parser.AST, errs *errorhandling.ValidationErrors) Visitor {
```
davenewza, line 13: We'd also want to validate the function's parameters (i.e. is a valid path provided and is the handler not conflicting with other function names)

jonbretman: Yeh I think there is some more validation to add, this is just a first pass.

The handler can conflict with action names though as these will be in a `routes` directory and not the `functions` directory.

PR #1743, opened 2025-02-26
fix: create hook and relationship reference

davenewza:
The table config generated for the functions runtime wasn't taking into consideration related models which may be multiple words strung together, e.g. `coAuthor`.  This fixes the issue experienced with create hooks failing when referencing a another model in the case of multiple words (coAuthor)

### DIFF
--- a/integration/testdata/functions_hooks/schema.keel
+++ b/integration/testdata/functions_hooks/schema.keel
@@ -1,6 +1,7 @@
 model Book {
     fields {
-        author Author?
+        author Author? @relation(books)
+        coAuthor Author?
         title Text
         published Boolean @default(false)
         cover File?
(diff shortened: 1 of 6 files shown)


RutZap (approved):

node/codegen.go
```diff
@@ -1157,7 +1157,7 @@ func writeTableConfig(w *codegen.Writer, models []*proto.Model) {
 				tableConfigMap[casing.ToSnake(model.Name)] = tableConfig
 			}
 
-			tableConfig[field.Name] = relationshipConfig
+			tableConfig[casing.ToSnake(field.Name)] = relationshipConfig
```
jonbretman, line 1160: wow I can't believe we didn't catch this before.

integration/testdata/functions_hooks/schema.keel
```diff
@@ -1,6 +1,7 @@
 model Book {
     fields {
-        author Author?
+        author Author? @relation(books)
+        coAuthor Author?
```
jonbretman, line 4: also amazed we had no integration tests with a multi-word field...

jonbretman (approved):

node/codegen.go
davenewza: Agreed!

PR #1750, opened 2025-02-28
build: goreleaser duplicate names

davenewza:
Fixes the goreleaser error encountered:

```
 error=archive named dist/keel_0.406.1_linux_amd64.tar.gz already exists. Check your archive name template
```


jonbretman (approved):
I knew something would be wrong with this 😄

PR #1756, opened 2025-03-04
feat: route functions

jonbretman:
See the integration tests for examples.


node/codegen.go
```diff
@@ -167,6 +168,29 @@ func generateSdkPackage(schema *proto.Schema, cfg *config.ProjectConfig) codegen
...
+	w.Dedent()
+	w.Writeln("}")
+
+	w.Writeln("export type RouteFunctionResponse = {")
+	w.Indent()
+	w.Writeln("body: string;")
+	w.Writeln("statusCode?: number;")
+	w.Writeln("headers?: {[key: string]: string};")
+	w.Dedent()
+	w.Writeln("}")
```
davenewza, line 191: Returning this structure does make sense, but what is a little awkward is that in custom functions it's also possible to change the response headers and status code but by using `ctx.response`.  So now we have two different ways.  Do you think it might make sense that we use `ctx.response` here as well and only return a body?

integration/testdata/route_functions/routes/postHandler.ts
```diff
@@ -0,0 +1,14 @@
+import { RouteFunction } from "@teamkeel/sdk";
+
+const handler: RouteFunction = async (request, ctx) => {
+  const body = JSON.parse(request.body);
+
+  return {
+    body: JSON.stringify({
```
davenewza, line 7: Just a thought, but what about the type for `body` be `string | object` and if it's an object then we perform the JSON serialisation?

integration/testdata/route_functions/schema.keel
```diff
@@ -0,0 +1,15 @@
+routes {
+    get("/get/route", getHandler)
+    get("/path/param/route/:foo", pathParamHandler)
+    post("/post/route", postHandler)
+    post("/raw/body/route", rawBodyHandler)
+    post("/headers/route", headersHandler)
+    post("/database/route", databaseHandler)
+    put("/status/route", statusHandler)
```
davenewza, line 8: I guess we need some schema validations on these?

davenewza, line 8: Since they all have the same TS signature, I guess it's possible to define multiple routes with the same path and even use the same handler in more than one route, e.g.
```
get("/myroute", getHandler)
post("/myroute", getHandler)
```

node/codegen.go
```diff
@@ -167,6 +168,29 @@ func generateSdkPackage(schema *proto.Schema, cfg *config.ProjectConfig) codegen
...
+	w.Writeln("path: string;")
+	w.Writeln("params: {[key: string]: string | undefined};")
+	w.Writeln("query: string;")
+	w.Writeln("headers: Headers;")
+	w.Dedent()
+	w.Writeln("}")
+
+	w.Writeln("export type RouteFunctionResponse = {")
+	w.Indent()
+	w.Writeln("body: string;")
```
davenewza, line 187: Maybe `body?: string` as there might not be a need to return a body

davenewza (approved):

node/codegen.go
jonbretman: Yeh I did think about this but decided I didn't want to compromise this API for the sake of same-ness.

This just feels a bit awkward...

```ts
ctx.response.headers.set("foo", "bar");

return {
  body: '...'
}
```

integration/testdata/route_functions/routes/postHandler.ts
jonbretman: Yeh we can do this but I've tried to go for a "no magic" approach for this feature to start with as it's quite a low-level thing and just want something that _definitely_ works.

integration/testdata/route_functions/schema.keel
jonbretman: Yeh some but I'm not sure there is much. Maybe just that it is a valid path?

jonbretman: Yep this will work fine.

davenewza: Yeah, a valid string path and a valid hander name

RutZap (approved):

PR #1760, opened 2025-03-05
fix: prevent embedding data when no results are available

RutZap:
When resolving a `get` action which has no results (i.e. item not found), we should not attempt to retrieve any embedded data

### DIFF
--- a/runtime/actions/get.go
+++ b/runtime/actions/get.go
@@ -63,7 +63,7 @@ func Get(scope *Scope, input map[string]any) (map[string]any, error) {
 	}
 
 	// if we have embedded data, let's resolve it
-	if len(scope.Action.ResponseEmbeds) > 0 {
+	if len(scope.Action.ResponseEmbeds) > 0 && len(res) > 0 {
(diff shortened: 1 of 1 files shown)


runtime/actions/get.go
```diff
@@ -63,7 +63,7 @@ func Get(scope *Scope, input map[string]any) (map[string]any, error) {
 	}
 
 	// if we have embedded data, let's resolve it
-	if len(scope.Action.ResponseEmbeds) > 0 {
+	if len(scope.Action.ResponseEmbeds) > 0 && len(res) > 0 {
```
jonbretman, line 66: It might be cleaner to short-circuit this function and return early if `res` is nil, rather than continuing. I don't think there is anything more to do if the record doesn't exist.

RutZap: @jonbretman good call. I've updated it now to return after authorisation so that we can return permission denied responses before any information leaking ones such as `item not found`

davenewza (approved):

PR #1768, opened 2025-03-11
chore: improve resilience with invalid embedded tool groups

RutZap:
When configuring embedded tools, we must handle invalid configurations gracefully by not applying them. Future validation will flag these configs but we must not panic when attempting to apply the user config

### DIFF
--- a/tools/config.go
+++ b/tools/config.go
@@ -601,12 +601,24 @@ func (cfg *ToolGroupLinkConfig) hasChanges() bool {
 	return len(cfg.ResponseOverrides) > 0 || cfg.isDeleted() || (cfg.ActionLink != nil && cfg.ActionLink.hasChanges())
 }
 
+func (cfg *ToolGroupLinkConfig) getToolLinkID() string {
+	if cfg == nil || cfg.ActionLink == nil {
+		return ""
+	}
+
+	return cfg.ActionLink.ToolID
+}
+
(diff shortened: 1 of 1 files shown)


tools/config.go
```diff
@@ -601,12 +601,24 @@ func (cfg *ToolGroupLinkConfig) hasChanges() bool {
...
+
 func (cfgs ToolGroupLinkConfigs) applyOn(links []*toolsproto.ToolGroup_GroupActionLink) []*toolsproto.ToolGroup_GroupActionLink {
 	newTools := []*toolsproto.ToolGroup_GroupActionLink{}
 
 	// add all configured links and new links. If links are deleted, they are skipped
 	for _, cfg := range cfgs {
-		if configured := cfg.applyOn(toolsproto.FindToolGroupLinkByToolID(links, cfg.ActionLink.ToolID)); configured != nil {
+		if cfg.getToolLinkID() == "" {
+			// skip embedding configs with invalid tools links
+			continue
```
davenewza, line 619: An invalid tool would be one in which the config is malformed?  I'm usually pro failing in cases like this, otherwise perhaps we're "hiding" the error

davenewza: What we also really need is tracing for the CLI

RutZap: I think we should raise errors on badly structured tool config as part of keel validate. This is within scope of the next piece of work that we'll be undertaking for this. For now though, I believe this is sufficient to continue with the in-repo tools migration.

davenewza (approved):

PR #1771, opened 2025-03-12
fix: relaxing enum casing

davenewza:
Enum values were restricted to being upper case camel, for e.g. `ComputerParts`.  This PR relaxed this rule and allows for any casing, numbers and the underscore.


RutZap (approved):

jonbretman:
You'll have to change the formatter too as that currently forces to camel - https://github.com/teamkeel/keel/blame/main/schema/format/format.go#L531C1-L531C43

PR #1780, opened 2025-03-20
feat: seed data

tomfrew:
* Supports seed data in `seed/*.sql`
* When running `keel run` and it creates a fresh db or `--reset` is used then it will apply the seed data after the migrations
* Adds a `keel seed` command to manually trigger the seeding
* Adds `keel seed snapshot` which writes the current state of your keel tables to `seed/snapshot.sql`

Snapshotting supports:
* Handling all the data types
* Handling array fields
* Don't snapshot computed fields
* Deferring the constraints when seeding
* Support for the keel storage table to be able to seed files locally


cmd/program/model.go
```diff
@@ -372,6 +386,47 @@ func (m *Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+				return m, SnapshotDatabase(m.ProjectDir, m.Schema, m.Database)
+			}
+
+		}
+
+		// If seed data flag is set, run seed data
+		if m.SeedData {
+			// If seed data directory exists, run seed data
+			seedDir := filepath.Join(m.ProjectDir, "seed")
+			if _, err := os.Stat(seedDir); !os.IsNotExist(err) {
```
RutZap, line 401:
```suggestion
			if _, err := os.Stat(seedDir); err == nil {
```

I think we need to attempt to seed only if there aren't any errors

migrations/snapshot.go
```diff
@@ -0,0 +1,400 @@
...
+						handler = &ArrayTypeHandler{}
+					} else {
+						handler = typeHandlers[field.Type.Type]
+					}
+				}
+
+				// If no specific handler, use default
+				if handler == nil {
+					handler = &DefaultTypeHandler{}
+				}
```
RutZap, line 379:
```suggestion
				handler := &DefaultTypeHandler{}
				if field != nil {
					if field.Type.Repeated {
						handler = &ArrayTypeHandler{}
					} else {
						handler = typeHandlers[field.Type.Type]
					}
				}
```

migrations/snapshot.go
```diff
@@ -0,0 +1,400 @@
...
+				columns = append(columns, col)
+				continue
+			}
+			field := proto.FindField(m.Schema.Models, model, casing.ToLowerCamel(col))
+			if field != nil && field.ComputedExpression == nil {
+				columns = append(columns, col)
+			}
+		}
+
+		if len(result.Rows) == 0 {
```
RutZap, line 316: I think we should move this if statement above the columns for loop; saves a few cpu cycles 😄

migrations/snapshot.go
```diff
@@ -0,0 +1,400 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/teamkeel/keel/casing"
+	"github.com/teamkeel/keel/proto"
+)
+
+type TypeHandler interface {
```
RutZap, line 14: I'm not sure we should be exporting this interface as I don't think it's used outside the package

migrations/snapshot.go
```diff
@@ -0,0 +1,400 @@
...
+		sortedModels = append(sortedModels, model)
+	}
+	for _, model := range models {
+		visit(model)
+	}
+
+	// Reverse the order to get tables with no dependencies first
+	for i, j := 0, len(sortedModels)-1; i < j; i, j = i+1, j-1 {
+		sortedModels[i], sortedModels[j] = sortedModels[j], sortedModels[i]
+	}
```
RutZap, line 270:
```suggestion
	// Sort models based on dependencies (tables with no dependencies first)
	sortedModels := append([]string{}, models)
	
	slices.SortFunc(sortedModels, func(a, b string) int {
		return len(tableDeps[a]) - len(tableDeps[b])
	})
```

migrations/snapshot.go
jonbretman: I don't think this is really an issue. For functions it's worth "hiding" stuff you don't want people to use, but not sure it matters for types/interfaces.

migrations/snapshot.go
```diff
@@ -0,0 +1,400 @@
...
+
+func (h *FileTypeHandler) ToSQL(value interface{}, field *proto.Field) (string, error) {
+	if value == nil {
+		return "NULL", nil
+	}
+
+	switch v := value.(type) {
+	case map[string]interface{}:
+		// Create ordered map for consistent output
+		orderedMap := make(map[string]interface{})
```
jonbretman, line 83: Maps in Go are unordered, so this isn't doing what the comment or variable name suggests.

migrations/snapshot.go
```diff
@@ -0,0 +1,400 @@
...
+	default:
+		return "", fmt.Errorf("unsupported type for File: %T", value)
+	}
+}
+
+type ArrayTypeHandler struct{}
+
+func (h *ArrayTypeHandler) ToSQL(value interface{}, field *proto.Field) (string, error) {
+	switch v := value.(type) {
+	case string:
```
jonbretman, line 110: If the field is an array why would it be sometimes a string and sometimes a slice?

migrations/snapshot.go
```diff
@@ -0,0 +1,400 @@
...
+
+func (m *Migrations) SnapshotDatabase(ctx context.Context) (string, error) {
+	ctx, span := tracer.Start(ctx, "Snapshot Database V2")
+	defer span.End()
+
+	var output strings.Builder
+	models := m.Schema.ModelNames()
+
+	output.WriteString("BEGIN;\n\n")
+	output.WriteString("SET CONSTRAINTS ALL DEFERRED;\n\n")
```
jonbretman, line 236: If you're doing the work to insert data in the right order I don't think you need to do this as well. Also I don't think this does anything unless the constraint itself was _created_ with `DEFERRABLE`:

> Currently, only UNIQUE, PRIMARY KEY, REFERENCES (foreign key), and EXCLUDE constraints are affected by this setting. NOT NULL and CHECK constraints are always checked immediately when a row is inserted or modified (not at the end of the statement). Uniqueness and exclusion constraints that have not been declared DEFERRABLE are also checked immediately.

https://www.postgresql.org/docs/current/sql-set-constraints.html

migrations/snapshot.go
jonbretman: AI: 0
Human: 1

😄

migrations/snapshot.go
```diff
@@ -0,0 +1,400 @@
...
+	for i, model := range sortedModels {
+		table := Identifier(model)
+
+		// Check if keel_storage table exists before querying
+		if model == "keel_storage" {
+			exists, err := m.database.ExecuteQuery(ctx, "SELECT EXISTS (SELECT FROM information_schema.tables WHERE table_name = 'keel_storage')")
+			if err != nil {
+				return "", fmt.Errorf("error checking if keel_storage table exists: %w", err)
+			}
+			if !exists.Rows[0]["exists"].(bool) {
```
jonbretman, line 291: Is there not an issue with casing here? In the query it's `EXISTS` but here you check `exists`. Or do column names come back always lowercase?

migrations/sql.go
```diff
@@ -173,7 +173,7 @@ func addColumnStmt(schema *proto.Schema, modelName string, field *proto.Field) (
 // addForeignKeyConstraintStmt generates a string of this form:
 // ALTER TABLE "thisTable" ADD FOREIGN KEY ("thisColumn") REFERENCES "otherTable"("otherColumn") ON DELETE CASCADE;
 func addForeignKeyConstraintStmt(thisTable string, thisColumn string, otherTable string, otherColumn string, onDelete string) string {
-	return fmt.Sprintf("ALTER TABLE %s ADD FOREIGN KEY (%s) REFERENCES %s(%s) ON DELETE %s;",
+	return fmt.Sprintf("ALTER TABLE %s ADD FOREIGN KEY (%s) REFERENCES %s(%s) ON DELETE %s DEFERRABLE INITIALLY IMMEDIATE;",
```
jonbretman, line 176: Ah I see you've added this. I suppose this is fine, but then if we're doing this maybe we don't need to faff around working out the right order to insert tables, which would simplify that code quite a bit.

davenewza (approved):

RutZap (approved):

PR #1785, opened 2025-03-27
chore: add db migration for flows tables

RutZap:


migrations/flows.sql
```diff
@@ -0,0 +1,23 @@
+CREATE TABLE IF NOT EXISTS keel_flow (
+	"id" text NOT NULL DEFAULT ksuid() PRIMARY KEY,
+	"name" TEXT NOT NULL,
+	"status" TEXT NOT NULL,
+	"type" TEXT NOT NULL,
```
davenewza, line 5: I guess a flow wouldn't have a `type` would it?

davenewza (approved):

migrations/flows.sql
RutZap: unless we use the type to store the invocation type (e.g. scheduled/manual)

PR #1786, opened 2025-03-27
feat: implement flows api

RutZap:


runtime/flows/scope.go
```diff
@@ -0,0 +1,33 @@
...
+type Scope struct {
+	Context context.Context
+	Flow    *proto.Flow
+	Schema  *proto.Schema
+}
+
+// WithContext sets the given context on the scope
+func (s *Scope) WithContext(ctx context.Context) *Scope {
+	return &Scope{
+		Context: ctx,
```
jonbretman, line 21: You shouldn't really keep `ctx` in a struct like this, better to always pass it as first argument. This is because the ctx changes throughout a program and tracing relies on using the right context, otherwise spans will be all messed up.

runtime/flows/start.go
```diff
@@ -0,0 +1,52 @@
...
+		jsonInputs = inputsMap
+	}
+
+	run := FlowRun{
+		Status: StatusNew,
+		Input:  &jsonInputs,
+		Name:   scope.Flow.Name,
+	}
+
+	database, err := db.GetDatabase(scope.Context)
```
jonbretman, line 36: This is a good example of why you shouldn't do this. Here you're passing `scope.Context` but actually you want the `ctx` variable created on L17.

runtime/flows/scope.go
RutZap: Indeed, I'm not a big fan of this approach, the only reason I've done it this way is to be consistent with the way we're handling actions in the runtime. If we're happy to move away from this approach I'll refactor 👍

davenewza: Yeah, I think we should move away from it.  Should we get this merged into the workflows branch so long though so that we can keep cracking on?

davenewza (approved):
LGTM, but let's rework that ctx stuff before going to main

PR #1795, opened 2025-04-02
chore: handle FlowRunUpdated responses from flow runtime

RutZap:


functions/functions.go
```diff
@@ -430,16 +430,16 @@ func CallFlow(ctx context.Context, flow *proto.Flow, runId string) error {
...
 
 	if resp.Error != nil {
 		span.SetStatus(codes.Error, resp.Error.Message)
 		span.SetAttributes(attribute.Int("error.code", int(resp.Error.Code)))
-		return toRuntimeError(resp.Error)
+		return nil, nil, toRuntimeError(resp.Error)
 	}
 
-	return nil
+	return resp.Result, resp.Meta, nil
```
davenewza, line 442: Not sure yet if we need the meta data

davenewza (approved):

PR #1796, opened 2025-04-02
feat: flows mvp

davenewza:
Flows MVP
===

This PR features our first pass of the upcoming flows feature.


cmd/program/views.go
```diff
@@ -121,6 +121,16 @@ func renderRun(m *Model) string {
...
 		b.WriteString(colors.Blue("https://console.keel.so/local").Highlight().String())
 		b.WriteString("\n")
+		if m.Schema.HasFlows() {
+			addr := fmt.Sprintf("http://localhost:%s", m.Port)
+			if m.CustomHostname != "" {
+				addr = m.CustomHostname
+			}
+
+			b.WriteString("Flows api: ")
+			b.WriteString(colors.Blue(fmt.Sprintf("%s/flows/json", addr)).Highlight().String())
```
jonbretman, line 131: I'm not sure it's very useful to output this, at least for now we're not expecting anyone to be consuming this are we?

packages/functions-runtime/src/routes/customResponse.ts
```diff
@@ -0,0 +1,16 @@
+import { RouteFunction } from "@teamkeel/sdk";
```
jonbretman, line 1: What's going on with this file?

testing/testing.go
```diff
@@ -275,6 +276,17 @@ func Run(ctx context.Context, opts *RunnerOpts) error {
 					return
 				}
 
+				writeJSON(w, http.StatusOK, map[string]any{})
+				return
+			case FlowsPath:
```
jonbretman, line 281: You shouldn't really have to do this as the runtime handles the routing. We needed this for subscribers/jobs because there is no API to execute them but we do have an API for running a flow.

packages/functions-runtime/src/routes/customResponse.ts
davenewza: AI shenanigans

cmd/program/views.go
davenewza: @RutZap @jonbretman Im going to remove this for now

testing/testing.go
davenewza: That's a good point.  We actually haven't done much on the integration testing side, so will leave this as is for now until we work on the test stuff

tomfrew (approved):

PR #1800, opened 2025-04-03
chore: update order of generated entry activity actions

RutZap:
Changing the order of generated entry activity actions by pushing Deletes to the end of the list


davenewza (approved):
LGTM

PR #1804, opened 2025-04-04
feat: flow orchestration via SQS

RutZap:
Refactors the flow orchestrator to handle events and trigger new orchestrations by pushing events for itself. In the case of the orchestrator being initialised without a SQS client & queue URL, the events will be handled synchroneously instead of being pushed onto a queue.

Keel deploy will now provision a new queue for flows, and a new runtime lambda in flow mode, which is invoked via queue events.

### DIFF
--- a/deploy/lambdas/runtime/flows.go
+++ b/deploy/lambdas/runtime/flows.go
@@ -2,41 +2,91 @@ package runtime
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 
-	"github.com/teamkeel/keel/runtime"
+	lambdaevents "github.com/aws/aws-lambda-go/events"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/sqs"
+	"github.com/sirupsen/logrus"
+	"github.com/teamkeel/keel/proto"
+	"github.com/teamkeel/keel/runtime/flows"
+	"github.com/teamkeel/keel/util"
+	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/trace"
 )
 
-type RunFlowPayload struct {
-	ID string `json:"id"`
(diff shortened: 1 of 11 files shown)


deploy/program.go
```diff
@@ -264,7 +272,8 @@ func createProgram(args *NewProgramArgs) pulumi.RunFunc {
 			"KEEL_SECRETS":           pulumi.String(strings.Join(secretNames, ":")),
 			"KEEL_FILES_BUCKET_NAME": filesBucket.Bucket,
 			"KEEL_FUNCTIONS_ARN":     functions.Arn,
-			"KEEL_QUEUE_URL":         queue.Url,
+			"KEEL_EVENTS_QUEUE_URL":  eventQueue.Url,
```
davenewza, line 275: Good rename 👍

deploy/lambdas/runtime/flows.go
```diff
@@ -2,41 +2,91 @@ package runtime
...
+	if err != nil {
+		return nil, err
+	}
+
+	opts := []func(*sqs.Options){}
+	if awsEndpoint != "" {
+		opts = append(opts, func(o *sqs.Options) {
+			o.BaseEndpoint = &awsEndpoint
+		})
+	}
```
davenewza, line 31: This means the orchestrator knows about SQS (and therefore dependent on it).  Im wondering if we could decouple them such that adding a new queue implementation wouldn't mean actually reworking the orchestrator.  I'm not massively phased by this now as maybe it's overengineering.

runtime/flows/events.go
```diff
@@ -1,10 +1,77 @@
 package flows
 
+import (
+	"encoding/json"
+	"fmt"
+)
+
+const (
+	EventNameFlowRunUpdated = "flow.updated"
```
davenewza, line 9: We could replace `flow.updated` with `flow.disrupted` (meaning we exited for some reason) and `flow.completed` (meaning the flow completed successfully)

davenewza (approved):

PR #1813, opened 2025-04-09
build: move dev server generation into deploy.Build function

jonbretman:
* Reduces duplication of dev server generation - this now only happens as part of calling `deploy.Build`
* Updates `generate`, `init`, and `run` CLI commands to use `deploy.Build`


davenewza (approved):
Lovely!

PR #1824, opened 2025-04-17
feat: set UI step value endpoint

RutZap:
Introduces a new workflow API endpoint:

* `PUT flows/json/[flowName]/[runID]/[stepID]`

This sets the given user input as the value of the given `PENDING` `UI` step. It marks the step as `COMPLETED` and then triggers the orchestrator to continue running the flow.

### DIFF
--- a/runtime/flows/flow.go
+++ b/runtime/flows/flow.go
@@ -20,24 +20,22 @@ var tracer = otel.Tracer("github.com/teamkeel/keel/runtime/flows")
 type Status string
 
 const (
-	StatusNew       Status = "new"
-	StatusRunning   Status = "running"
-	StatusWaiting   Status = "waiting"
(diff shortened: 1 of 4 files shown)


runtime/flows/flow.go
```diff
@@ -20,24 +20,22 @@ var tracer = otel.Tracer("github.com/teamkeel/keel/runtime/flows")
 type Status string
 
 const (
-	StatusNew       Status = "new"
-	StatusRunning   Status = "running"
-	StatusWaiting   Status = "waiting"
-	StatusFailed    Status = "failed"
-	StatusCompleted Status = "completed"
+	StatusNew       Status = "NEW"
```
davenewza, line 23: Perhaps these could be renamed to FlowStatusNew, etc.

davenewza (approved):

PR #1832, opened 2025-04-24
feat: add openapi.json endpoint to the flows API

RutZap:
Adds a new endpoint: `GET /flows/json/openapi.json` 

Generates a openapi schema with the input messages used to start each flow.

Refactored the openapi test to assert either job schemas, flow schemas or general API schemas

### DIFF
--- a/runtime/openapi/openapi.go
+++ b/runtime/openapi/openapi.go
@@ -3,6 +3,7 @@ package openapi
 import (
 	"context"
 	"fmt"
+	"maps"
 	"strings"
 
 	"github.com/teamkeel/keel/proto"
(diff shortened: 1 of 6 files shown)


runtime/openapi/openapi.go
```diff
@@ -237,3 +238,100 @@ func GenerateJob(ctx context.Context, schema *proto.Schema, jobName string) Open
 
 	return spec
 }
+
+// GenerateFlows generates an openAPI schema for the Flows API for the given schema
+func GenerateFlows(ctx context.Context, schema *proto.Schema) OpenAPI {
+	runResponseSchema := jsonschema.JSONSchema{
```
jonbretman, line 244: since this stuff is all static maybe move it up to package scope rather than in the function?

jonbretman: Just seen that we mutate the spec later on so don't move it 😄

PR #1833, opened 2025-04-24
fix: async local flows

RutZap:


jonbretman (approved):
My only worry is that this won't play nicely with bubbletea for some reason, but worth a try I think.

PR #1837, opened 2025-04-25
fix: store traceIDs with flow runs

RutZap:
* Adds a new column to the keel flow runs db: traceID
* Sets the traceID to the flow run at point of creation

### DIFF
--- a/migrations/flows.sql
+++ b/migrations/flows.sql
@@ -1,6 +1,7 @@
 CREATE TABLE IF NOT EXISTS keel_flow_run (
 	"id" text NOT NULL DEFAULT ksuid() PRIMARY KEY,
 	"name" TEXT NOT NULL,
+	"trace_id" TEXT,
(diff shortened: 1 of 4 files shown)


migrations/flows.sql
```diff
@@ -1,6 +1,7 @@
 CREATE TABLE IF NOT EXISTS keel_flow_run (
 	"id" text NOT NULL DEFAULT ksuid() PRIMARY KEY,
 	"name" TEXT NOT NULL,
+	"trace_id" TEXT,
```
jonbretman, line 4: I guess this is fine for now but we will need to think about how to keep these tables updates if stuff changes. We _could_ consider using the proto schema to create these tables, like we do with Identity.

RutZap: Yeah it's not ideal. The difference we have between these runs/steps and audit/identity is that they're not built in keel models. I mean, they could be but they won't necessarily be very useful.

PR #1838, opened 2025-04-28
fix: flow definition and stages first pass

davenewza:
The flow definition is now improved to include the flow config and the function argument is correctly typed with `StepContext`:

```ts
export declare const MyFlow: {
  <C extends runtime.FlowConfig>(
    config: C,
    fn: (ctx: runtime.StepContext<C>, inputs: MyFlowMessage) => Promise<void>
  ): {
    config: C;
    fn: (ctx: runtime.StepContext<C>, inputs: MyFlowMessage) => Promise<void>;
  };
};
```

But, for some reason the stage constraints aren't being passed down into the UI stage types correctly.

The flow config is being returned from the functions runtime now:

image


RutZap:
Since we're changing the definition of the flow functions, I think we'll need to change the scaffolding as well in `node/scaffold.go`

node/scaffold_test.go
```diff
@@ -191,6 +197,34 @@ export default DoSomethingElse(async (ctx, event) => {
...
 		},
+		&codegen.GeneratedFile{
+			Contents: `
+import { MyFlow, FlowConfig } from '@teamkeel/sdk';
+
+const config: FlowConfig = {
+	// See https://docs.keel.so/flows for options
+};
+
+// To learn more about flows, visit https://docs.keel.so/flows
```
jonbretman, line 208: Personally I find these comments linking to the docs quite annoying. Every time I scaffold something I just start by deleting them. Maybe it would make more sense to be part of the `keel generate` command rather than in the generated output?

davenewza: I also agree.  Made a ticket for this:  https://linear.app/keel/issue/KE-2570/move-scaffolded-function-comments-to-the-cli-output

PR #1865, opened 2025-05-05
fix: enum literal values in computed expressions

davenewza:
Fixing the use of enum literal values in computed field expressions.

### DIFF
--- a/runtime/actions/generate_computed.go
+++ b/runtime/actions/generate_computed.go
@@ -122,6 +122,29 @@ func (v *computedQueryGen) VisitLiteral(value any) error {
 
 func (v *computedQueryGen) VisitIdent(ident *parser.ExpressionIdent) error {
 	model := v.schema.FindModel(strcase.ToCamel(ident.Fragments[0]))
+
+	if model == nil {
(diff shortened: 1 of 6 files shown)


runtime/actions/generate_computed.go
```diff
@@ -122,6 +122,29 @@ func (v *computedQueryGen) VisitLiteral(value any) error {
...
+
+	if model == nil {
+		enum := v.schema.FindEnum(ident.Fragments[0])
+		if enum == nil {
+			return fmt.Errorf("model or enum not found: %s", ident.Fragments[0])
+		}
+
+		var value string
+		for _, v := range enum.Values {
+			if v.Name == ident.Fragments[1] {
```
jonbretman, line 134: This will panic if you use a bare enum as there won't be a second fragment. I would check that and return an error.

davenewza: Schema validation would have picked that up though.  I'm often a bit unsure about situations like these, but I think we once did decide to assume a valid schema structure beyond validations. @jonbretman

jonbretman (approved):

PR #1872, opened 2025-05-07
fix: flow step errors, timeouts and retry handling

davenewza:
`error` column
--
If an error occurs on a step, this will be stored in the new `error` column.

Timeouts and retries handling
--
These columns have been removed from the database, and now the functions runtime handles them.

### DIFF
--- a/runtime/flows/orchestrator.go
+++ b/runtime/flows/orchestrator.go
@@ -113,11 +113,6 @@ func (o *Orchestrator) orchestrateRun(ctx context.Context, runID string, data ma
 			return err
 		}
 
-		stepsMap := map[string][]Step{}
-		for _, step := range run.Steps {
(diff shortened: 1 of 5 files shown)


runtime/flows/orchestrator.go
```diff
@@ -116,15 +118,6 @@ func (o *Orchestrator) orchestrateRun(ctx context.Context, runID string, data ma
 			stepsMap[step.Name] = append(stepsMap[step.Name], step)
 		}
 
-		// Check to see if the retries have been exceeded
-		if len(run.Steps) > 0 {
```
RutZap, line 120: If we're removing this from the orchestrator, we can remove the `stepsMap` above as well

davenewza: Good point

PR #1873, opened 2025-05-07
chore: add a AWAITING_INPUT run status

RutZap:
Adds a new flow run status: `AWAITING_INPUT`. Whenever a flow run is orchestrated and it is pending a UI step, the status will be set to awaiting input. Once a flow is orchestrated again, with data when in an `AWAITING_INPUT` status, the status will be updated to `RUNNING` and the flow runtime invoked. 

This will allow FE to differentiate between flow runs that are currently in progress (executing) and the ones that are awaiting user input.


jonbretman (commented):
I'm not sure about this. Thinking about it now I think this makes more sense as a run status. Having the status and then a separate flag feels a bit messy.

PR #1874, opened 2025-05-07
chore: element testing

tomfrew:
* Creates `_testFlow` and `_testFlowContext` in `testingUtils.ts` to enable easy testing
* Adds some test for elements. 
* `[element].test.ts` for behaviour tests
* `[element].test-d.ts` for type tests

### DIFF
--- a/packages/functions-runtime/src/flows/testingUtils.ts
+++ b/packages/functions-runtime/src/flows/testingUtils.ts
@@ -1,0 +1,12 @@
+import { createFlowContext, FlowFunction } from ".";
+import { FlowConfig } from ".";
+
+export const _testFlowContext = <T extends FlowConfig>(config?: T) =>
+  createFlowContext<T>("test-run-id", {}, "test-span-id");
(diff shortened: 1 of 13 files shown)


packages/functions-runtime/src/flows/testingUtils.ts
```diff
@@ -0,0 +1,12 @@
+import { createFlowContext, FlowFunction } from ".";
+import { FlowConfig } from ".";
+
+export const _testFlowContext = <T extends FlowConfig>(config?: T) =>
+  createFlowContext<T>("test-run-id", {}, "test-span-id");
+
+export const _testFlow = <const C extends FlowConfig>(
```
jonbretman, line 7: why the leading underscores?

tomfrew: I went a bit back and forth on this. Just wanted to really make it clear they were internal testing only things. Might might be overkill

PR #1875, opened 2025-05-07
feat: add startTime and endTime for flow steps

RutZap:
Adds two new fields to flow steps: `startTime` and `endTime`. These fields are populated with the timestamp of when the step started execution, and respectively when it finished the execution. 

These values can be used to calculate the duration of that particular step.


runtime/flows/flow.go
```diff
@@ -87,17 +87,17 @@ func (r *Run) SetUIComponent(ui *JSONB) {
 }
 
 type Step struct {
-	ID          string     `json:"id" gorm:"primaryKey;not null;default:null"`
-	Name        string     `json:"name"`
-	RunID       string     `json:"runId"`
-	Status      StepStatus `json:"status"`
-	Type        StepType   `json:"type"`
-	Value       *JSONB     `json:"value" gorm:"type:jsonb"`
-	MaxRetries  int        `json:"max_retries"`
```
jonbretman, line 96: Did you mean to remove these? Do we not need them?

davenewza (approved):

runtime/flows/flow.go
davenewza: Yeah it was removed in a previous PR (as we discussed IRL)

PR #1876, opened 2025-05-07
feat: page implementation and UI OAS gen

tomfrew:
* Adds implementation of page which takes the options and returns the data for the UI API
* Adds JSON schema generation from types which is then imported into the flows OAS
* Adds all the flows apis to the OAS

**N.B.** We probably want to add the OAS generation into one of the make files


jonbretman:
Wouldn't we always want to run `generate-openapi` whenever we run `build`? If so you could make it a `postbuild` ?

PR #1878, opened 2025-05-08
test: more tests for flows

jonbretman:
Mostly just making the assertions on the API responses more robust. It does make the tests more verbose but for now I think we will benefit from knowing exactly what is being returned.

### DIFF
--- a/integration/testdata/flows/tests.test.ts
+++ b/integration/testdata/flows/tests.test.ts
@@ -37,7 +37,19 @@ test("flows - stepless flow", async () => {
     id: body.id,
     token,
   });
-  expect(flow.status).toBe("COMPLETED");
+
+  // Flow has no steps so should be synchronously completed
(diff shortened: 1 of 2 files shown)


davenewza (approved):

integration/testdata/flows/tests.test.ts
```diff
@@ -54,20 +66,68 @@ test("flows - first step is a function", async () => {
...
-  expect(body.steps[0].output).toBeUndefined();
+  // First step is a function so should be in status NEW - it will get run async via the queue
+  expect(body).toEqual({
+    id: expect.any(String),
+    input: {},
+    name: "SingleStep",
+    status: "RUNNING",
+    steps: [
+      {
+        id: expect.any(String),
```
davenewza, line 77: Oooh, this is handy!

PR #1879, opened 2025-05-09
fix: flow step errors, timeouts, retry handling and tests

davenewza:
**Fixes, improvements and tests for when an error is thrown in a flow step function.**

- Step retries were not triggering, so this has now been fixed.
- Returning the step `error` column in the API responses.
- Removed `max_retries` and `timeout_in_ms` from the API responses.
- Tests for step errors, retries and timeouts.

### DIFF
--- a/integration/testdata/flows/tests.test.ts
+++ b/integration/testdata/flows/tests.test.ts
@@ -11,11 +11,11 @@ TEST CASES
 [ ] Flow function with consecutive function steps
 [x] Flow function with alternating function and UI steps
 [ ] Error thrown in flow function
-[ ] Error thrown in step function
+[x] Error thrown in step function
 [ ] Step returning scalar value
-[ ] Step function retrying  
+[x] Step function retrying  
 [ ] Step function timing out 
-[ ] UI step validation
+[x] UI step validation
 [ ] Check full API responses
 [ ] All UI elements response
 [ ] Test all Keel types as inputs
@@ -77,8 +77,7 @@ test("flows - first step is a function", async () => {
         id: expect.any(String),
         runId: body.id,
         name: "insert thing",
-        max_retries: 0,
-        timeout_in_ms: 0,
+        error: null,
         ui: null,
         status: "NEW",
         type: "FUNCTION",
@@ -111,8 +110,7 @@ test("flows - first step is a function", async () => {
         id: body.steps[0].id,
         runId: body.id,
         name: "insert thing",
-        max_retries: 0,
-        timeout_in_ms: 0,
+        error: null,
         ui: null,
         status: "COMPLETED",
         type: "FUNCTION",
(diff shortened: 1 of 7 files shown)


packages/functions-runtime/src/handleFlow.js
```diff
@@ -123,11 +122,11 @@ async function handleFlow(request, config) {
           );
         }
 
-        return createJSONRPCSuccessResponse(request.id, {
```
RutZap, line 126: So if an unexpected error occurs when running a step, by returning an error here would it not make the orchestrator to not attempt to continue the flow, therefore not retrying the execution of the step?

davenewza: So this is not where step errors are handled.  This is when the an exception gets thrown in the flow body itself

integration/testdata/flows/tests.test.ts
```diff
@@ -353,13 +359,209 @@ test("flows - alternating step types", async () => {
...
+});
+
 test("flows - authorised starting, getting and listing flows", async () => {
   const adminToken = await getToken({ email: "admin@keel.xyz" });
   const userToken = await getToken({ email: "user@gmail.com" });
 
   const resListAdmin = await listFlows({ token: adminToken });
   expect(resListAdmin.status).toBe(200);
-  expect(resListAdmin.body.flows.length).toBe(3);
+  expect(resListAdmin.body.flows.length).toBe(5);
```
jonbretman, line 564: Is 5 the default number of retries?

packages/functions-runtime/src/handleFlow.js
jonbretman: I can't see a test for an error being thrown in the function body - I think worth adding one

RutZap: I think we not need to add a special case in here for `StepErrorDisrupt`?; i.e.:

```
if (e instanceof StepErrorDisrupt) {
      return createJSONRPCSuccessResponse(request.id, {
             runId: runId,
             runCompleted: false,
             config: flowConfig,
        });
```

so then the flow can be retried until max retries is reached.

integration/testdata/flows/tests.test.ts
davenewza: @jonbretman At the moment, yes

packages/functions-runtime/src/handleFlow.js
davenewza: > I can't see a test for an error being thrown in the function body - I think worth adding one

Yeah, I'll do that in a new PR as I have a few questions

RutZap (approved):

packages/functions-runtime/src/handleFlow.js
davenewza: > @davenewza sorry, ignore my previous comment, I've just realised that we're no longer throwing the `StepErrorDisrupt` but instead we create a new step and throw `StepCreatedDisrupt`

Yes, that's right

PR #1888, opened 2025-05-12
test: add new test case for flows

RutZap:
Adds a test case for consecutive function steps with stages config


integration/testdata/flows/tests.test.ts
```diff
@@ -67,6 +67,118 @@ test("flows - scalar step", async () => {
...
+  expect(body).toEqual({
+    id: expect.any(String),
+    traceId: expect.any(String),
+    status: "RUNNING",
+    name: "OnlyFunctions",
+    input: {
+      name: "My Thing",
+      age: 25,
+    },
+    steps: [
```
davenewza, line 92: Seems we aren't returning the `stage` column from the API yet.  We should be testing that here

PR #1898, opened 2025-05-14
fix: normalise flow config

tomfrew:
* Normalise the flow config so we return a most consistent structure over the api 
* Stages are always object format
* Default to flow name if there isn't a title set


davenewza (approved):
LGTM

PR #1941, opened 2025-06-18
fix: add identity to flow context

RutZap:
`ctx.identity` will now be available within flow functions and steps.

### DIFF
--- a/packages/functions-runtime/src/flows/index.ts
+++ b/packages/functions-runtime/src/flows/index.ts
@@ -39,7 +39,7 @@ const defaultOpts = {
   timeout: 60000,
 };
 
-export interface FlowContext<C extends FlowConfig, E = any, S = any> {
+export interface FlowContext<C extends FlowConfig, E = any, S = any, Id = any> {
   // Defines a function step that will be run in the flow.
   step: Step<C>;
   // Defines a UI step that will be run in the flow.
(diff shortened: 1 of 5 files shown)


packages/functions-runtime/src/flows/index.ts
```diff
@@ -39,14 +39,15 @@ const defaultOpts = {
   timeout: 60000,
 };
 
-export interface FlowContext<C extends FlowConfig, E = any, S = any> {
+export interface FlowContext<C extends FlowConfig, E = any, S = any, Id = any> {
```
davenewza, line 42: I think we can do this because these types will and must always be provided

```suggestion
export interface FlowContext<C extends FlowConfig, E, S, Id> {
```

packages/functions-runtime/src/flows/index.ts
```diff
@@ -114,8 +115,9 @@ export type FlowFunction<
   C extends FlowConfig,
   E extends any = {},
   S extends any = {},
+  Id extends any = {},
   I extends any = {},
-> = (ctx: FlowContext<C, E, S>, inputs: I) => Promise<void>;
+> = (ctx: FlowContext<C, E, S, Id>, inputs: I) => Promise<void>;
```
davenewza, line 120: I wonder if `Id` is perhaps a little misleading as it makes me think of an `id` field and not the identity model

PR #1954, opened 2025-06-26
fix: persisting flows config

davenewza:
We now persist the flows config so that it is available without having to run the actual flow function.  This is particularly important for the final completion step which needs to be redisplayable.

### DIFF
--- a/runtime/flows/orchestrator.go
+++ b/runtime/flows/orchestrator.go
@@ -143,7 +143,7 @@ func (o *Orchestrator) orchestrateRun(ctx context.Context, runID string, inputs 
 				return err, resp.GetUIComponents()
 			}
 
-			_, err = completeRun(ctx, run.ID, resp.Data, resp.Config)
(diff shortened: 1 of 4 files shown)


runtime/flows/orchestrator.go
```diff
@@ -155,10 +155,16 @@ func (o *Orchestrator) orchestrateRun(ctx context.Context, runID string, inputs
 
 		// Check to see if we're in a Pending UI step, break orchestration
 		if run.HasPendingUIStep() {
-			_, err = updateRun(ctx, run.ID, StatusAwaitingInput, nil)
+			_, err = updateRun(ctx, run.ID, StatusAwaitingInput, resp.Config)
 			return err, resp.GetUIComponents()
 		}
 
+		// Set the config
+		_, err = updateRun(ctx, run.ID, run.Status, resp.Config)
```
RutZap, line 163: would it not be a small improvement here if we check if the `resp.config` differs from `run.Config` and only update it if anything changed?

RutZap (approved):

PR #1960, opened 2025-07-03
fix: flow step args, onfailure handler and noretryable errors

davenewza:
Steps are now provides with an argument object which consists of `attempt` (the function executation attempt number, start from 0) and `config` (the configuration object passed into the step).

An `onFailure` callback handler can be provided. This will only be executed once all retries have been exhausted.

A `NonRetryableError` can be thrown in the step function to bypass any retrying logic.


packages/functions-runtime/src/flows/index.ts
```diff
@@ -67,40 +68,40 @@ type JsonSerializable =
   | JsonSerializable[]
   | { [key: string]: JsonSerializable };
 
+type StepOptions<C extends FlowConfig> = {
+  stage?: ExtractStageKeys<C>;
+  /** Number of times to retry the step after it fails. Defaults to 4. */
+  retries?: number;
+  /** Maximum time in milliseconds to wait for the step to complete. Defaults to 60000 (1 minute). */
+  timeout?: number;
+  /** A function to call if the step fails after exhausts all it retr. */
```
RutZap, line 77:
```suggestion
  /** A function to call if the step fails after it exhausts all retries. */
```

RutZap (approved):

PR #1961, opened 2025-07-04
fix: flows page validation

davenewza:
Validation for pages.  I've also changed the validation functions so that you can also return `undefined` which means you can leave out the return statements entirely which feels good to me.  See example:

```ts
await ctx.ui.page("first page", {
    content: [
      ctx.ui.inputs.text("email", {
        label: "Email",
        optional: true,
        validate(value) {
          if (!emailRegex.test(value)) {
            return "Not a valid email";
          }
        },
      }),
      ctx.ui.inputs.text("phone", {
        label: "Phone",
        optional: true,
        validate(value) {
          if (!phoneRegex.test(value)) {
            return "Not a valid phone number";
          }
        },
      }),
    ],
    validate: async (data) => {
      if (!data.email && !data.phone) {
        return "Email or phone is required";
      }
    },
  });
```

### DIFF
--- a/integration/testdata/flows/tests.test.ts
+++ b/integration/testdata/flows/tests.test.ts
@@ -855,6 +855,104 @@ test("flows - boolean input validation", async () => {
   });
 });
 
+test.only("flows - page validation", async () => {
+  const token = await getToken({ email: "admin@keel.xyz" });
+  let { status, body } = await startFlow({
+    name: "ValidationPage",
+    token,
+    body: {},
+  });
+
+  expect(status).toBe(200);
+  expect(body.steps[0].status).toBe("PENDING");
+
+  const runId = body.id;
+  let stepId = body.steps[0].id;
+
+  ({ status, body } = await putStepValues({
+    name: "ValidationPage",
+    runId,
+    stepId,
+    token,
+    values: {},
+    action: null,
+  }));
+
+  expect(status).toBe(200);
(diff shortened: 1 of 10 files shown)


integration/testdata/flows/tests.test.ts
```diff
@@ -855,6 +855,104 @@ test("flows - boolean input validation", async () => {
   });
 });
 
+test.only("flows - page validation", async () => {
```
RutZap, line 858:
```suggestion
test("flows - page validation", async () => {
```

packages/functions-runtime/src/flows/ui/page.ts
```diff
@@ -101,7 +103,13 @@ export async function page<
       .filter(Boolean)
   )) as UiElementApiResponses;
 
-  // TODO support the page level validation function
+  if (data && options.validate) {
+    const validationResult = await options.validate(data);
```
RutZap, line 107: Should we wrap this in a try catch to account for failed validation functions?

integration/testdata/flows/tests.test.ts
davenewza: Oh my, thanks!

packages/functions-runtime/src/flows/ui/page.ts
davenewza: @RutZap yeah, so we need to figure out how to handle user code errors

PR #1977, opened 2025-07-23
fix: parse @schedule attributes for flows

RutZap:
Adds support for parsing `@schedule` attributes on flows.

Adds validation ensuring that scheduled flows cannot have inputs.


davenewza (approved):
LGTM!

PR #1978, opened 2025-07-23
fix: abstract the sqs event sender from the orchestrator

RutZap:
Updates the flows orchestrator to receive a EventSender instead of a sqs queue & sqs client.

Introduces a SQSEventSender & a NoQueueEventSender (instead of sending the message to the orchestrator via the queue, it invokes the orchestrator's event handler in a new goroutine).

### DIFF
--- a/cmd/program/model.go
+++ b/cmd/program/model.go
@@ -591,7 +591,7 @@ func (m *Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		}
 
 		// Setting the flows orchestrator
-		ctx = flows.WithOrchestrator(ctx, flows.NewOrchestrator(m.Schema))
+		ctx = flows.WithOrchestrator(ctx, flows.NewOrchestrator(m.Schema, flows.WithNoQueueEventSender()))
 
 		r = msg.r.WithContext(ctx)
 		m.RuntimeHandler.ServeHTTP(msg.w, r)
--- a/runtime/flows/event_sender.go
+++ b/runtime/flows/event_sender.go
@@ -1,0 +1,64 @@
+package flows
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/aws/aws-sdk-go-v2/service/sqs"
+	"github.com/aws/aws-sdk-go/aws"
+)
+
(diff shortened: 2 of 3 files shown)


runtime/flows/event_sender.go
```diff
@@ -0,0 +1,64 @@
...
+var _ EventSender = &NoQueueEventSender{}
+
+func NewNoQueueEventSender(o *Orchestrator) *NoQueueEventSender {
+	return &NoQueueEventSender{
+		orchestrator: o,
+	}
+}
+
+func (s *NoQueueEventSender) Send(ctx context.Context, payload *EventWrapper) error {
+	go s.orchestrator.HandleEvent(ctx, payload) //nolint we're "simulating" an async queue
```
davenewza, line 61: whoop

cmd/program/model.go
```diff
@@ -591,7 +591,7 @@ func (m *Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		}
 
 		// Setting the flows orchestrator
-		ctx = flows.WithOrchestrator(ctx, flows.NewOrchestrator(m.Schema))
+		ctx = flows.WithOrchestrator(ctx, flows.NewOrchestrator(m.Schema, flows.WithNoQueueEventSender()))
```
davenewza, line 594: Out of interest, do we even need the `flows.NewOrchestrator` here?  Is it serving a purpose when setting up flows?  Maybe we could simplify this to `flows.WithOrchestrator(m.Schema, flows.WithNoQueueEventSender)`?

runtime/flows/event_sender.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/aws/aws-sdk-go-v2/service/sqs"
+	"github.com/aws/aws-sdk-go/aws"
+)
+
+type EventSender interface {
```
davenewza, line 11: I wonder if we should quality this name further to differentiate with keel events.  Maybe FlowEventSender?

davenewza: Very good point

PR #1979, opened 2025-07-24
fix: support running scheduled flows locally

RutZap:
This PR adds support for running scheduled flows in a local environment i.e. `keel run`.

After starting the functions server, we start a cron server which executes given functions according to a cron schedule. The functions executed in this instance will be simulate a `FlowRunStarted` event being placed on the orchestrator's queue; then the orchestrator will handle the event, create a new flow run and then the normal flow of execution takes place. 

Since the functions server is restarted/updated every time the schema changes, this resets the cron jobs thus ensuring that schedule updates are reflected accordingly

### DIFF
--- a/cmd/program/commands.go
+++ b/cmd/program/commands.go
@@ -21,6 +21,7 @@ import (
 	"github.com/Masterminds/semver/v3"
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/fsnotify/fsnotify"
+	"github.com/robfig/cron/v3"
 	"github.com/teamkeel/keel/cmd/cliconfig"
 	"github.com/teamkeel/keel/cmd/database"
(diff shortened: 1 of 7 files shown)


cmd/program/commands.go
```diff
@@ -21,16 +21,20 @@ import (
 	"github.com/Masterminds/semver/v3"
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/fsnotify/fsnotify"
+	"github.com/robfig/cron/v3"
```
davenewza, line 24: We might need to check that this cron format is compatible with ours (mostly).  Ours is the same with eventbridge.  Sadly, there are a few cron formats floating around

davenewza (approved):

PR #1987, opened 2025-08-01
feat: flows testing sdk

davenewza:
Testing SDK for flows.  Like the other testing SDKs, this is imported as a module into your test file:

```ts
import { flows } from "@teamkeel/testing";
```

Every flow will then be available from that modules.  For example, to start a new flow:

```ts
const flow = await flows.onboardUser.withToken(token).start({ name: "Keelson" });
```

There are similar functions for `get`, `cancel`, `untilFinished` and `untilAwaitingInput`.


packages/testing-runtime/src/parsing.mjs
```diff
@@ -0,0 +1,114 @@
...
+        row[key] = parseOutputs(value);
+      }
+    } else if (
+      Array.isArray(value) &&
+      value.every((item) => typeof item === "object" && item !== null)
+    ) {
+      const arr = [];
+      for (let item of value) {
+        if (item.key && item.size && item.filename && item.contentType) {
+          arr.push(File.fromDbRecord(item));
```
RutZap, line 61: We need to import `File` for this.

davenewza: Oops!

PR #2006, opened 2025-09-16
chore: basic task assignment

davenewza:
This performs basic task assignment when calling upon the `POST topics/{name}/tasks/next` endpoint.  The next available task will be assigned to the currently authenticated user.  The following basic rules apply:
 - If the user already has a task assigned to them, then return that task instead
 - The candidate task is the next NEW and unassigned task whose deferred date has passed (if even set)

Still to do is to apply task ordering by the `@order` attribute.  Also need to write a lot more tests.

### DIFF
--- a/integration/testdata/tasks/tests.test.ts
+++ b/integration/testdata/tasks/tests.test.ts
@@ -118,7 +118,7 @@ async function getToken({ email }) {
 }
 
 async function createTask({ topic, token }) {
-  const url = `${process.env.KEEL_TESTING_API_URL}/topics/${topic}/tasks`;
+  const url = `${process.env.KEEL_TESTING_API_URL}/topics/json/${topic}/tasks`;
   const res = await fetch(url, {
     method: "POST",
     headers: {
--- a/integration/testdata/tasks/tests.test.ts
+++ b/integration/testdata/tasks/tests.test.ts
@@ -151,7 +151,7 @@ async function listTasks({ topic, token }) {
 }
 
 async function nextTask({ topic, token }) {
-  const url = `${process.env.KEEL_TESTING_API_URL}/topics/${topic}/tasks/next`;
+  const url = `${process.env.KEEL_TESTING_API_URL}/topics/json/${topic}/tasks/next`;
   const res = await fetch(url, {
     method: "POST",
     headers: {
--- a/integration/testdata/tasks/tests.test.ts
+++ b/integration/testdata/tasks/tests.test.ts
@@ -34,7 +34,10 @@ test("tasks - next - no tasks exist", async () => {
   const token = await getToken({ email: "admin@keel.xyz" });
   const res = await nextTask({ topic: "DispatchOrder", token: token });
   expect(res.status).toBe(404);
-  expect(res.body).toEqual({});
+  expect(res.body).toEqual({
+    code: "ERR_RECORD_NOT_FOUND",
+    message: "Not found",
+  });
 });
 
 test("tasks - next - successfully assigned", async () => {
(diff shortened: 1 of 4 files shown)


runtime/tasks/tasks.go
```diff
@@ -460,3 +462,95 @@ func AssignTask(ctx context.Context, pbTask *proto.Task, id string, assignedTo,
...
+			span.RecordError(err, trace.WithStackTrace(true))
+			span.SetStatus(codes.Error, err.Error())
+		}
+	}()
+
+	identity, err := auth.GetIdentity(ctx)
+	if err != nil {
+		return nil, err
+	}
+	identityID, _ := identity[parser.FieldNameId].(string)
```
RutZap, line 482: We already do this in the handler, I think it may be worth sending the `identityID` as an arg to this method; so then it's agnostic of what's set in the context

runtime/apis/tasksapi/handler.go
```diff
@@ -131,8 +131,25 @@ func Handler(s *proto.Schema) common.HandlerFunc {
...
+			switch pathParts[2] {
+			case "next":
+				if r.Method != http.MethodPost {
+					return httpjson.NewErrorResponse(ctx, common.NewHttpMethodNotAllowedError("only HTTP POST accepted"), nil)
+				}
+
+				task, err := tasks.NextTask(ctx, identityID)
+				if err != nil {
+					if errors.Is(err, tasks.ErrTaskNotFound) {
+						return common.NewJsonResponse(http.StatusNotFound, map[string]any{}, nil)
```
RutZap, line 143:
```suggestion
						return httpjson.NewErrorResponse(ctx, common.NewNotFoundError("Not found"), nil)
```
This should maintain the error structure format; e,.g.:

```
{
    "code": "ERR_RECORD_NOT_FOUND",
    "message": "Not found"
}```

File: integration/testdata/tasks/tests.test.ts [hunk 3]
```diff
@@ -0,0 +1,167 @@
...
+  });
+
+  return {
+    status: res.status,
+    body: await res.json(),
+  };
+}
+
+async function nextTask({ topic, token }) {
+  const url = `${process.env.KEEL_TESTING_API_URL}/topics/${topic}/tasks/next`;
```
RutZap, line 154:
```suggestion
  const url = `${process.env.KEEL_TESTING_API_URL}/topics/json/${topic}/tasks/next`;
```

integration/testdata/tasks/tests.test.ts
```diff
@@ -0,0 +1,167 @@
...
+  });
+
+  return {
+    status: res.status,
+    body: await res.json(),
+  };
+}
+
+async function listTasks({ topic, token }) {
+  const url = `${process.env.KEEL_TESTING_API_URL}/topics/${topic}/tasks`;
```
RutZap, line 137:
```suggestion
  const url = `${process.env.KEEL_TESTING_API_URL}/topics/json/${topic}/tasks`;
```

integration/testdata/tasks/tests.test.ts
```diff
@@ -0,0 +1,167 @@
...
+    {
+      emailVerified: true,
+    }
+  );
+
+  return token;
+}
+
+async function createTask({ topic, token }) {
+  const url = `${process.env.KEEL_TESTING_API_URL}/topics/${topic}/tasks`;
```
RutZap, line 121:
```suggestion
  const url = `${process.env.KEEL_TESTING_API_URL}/topics/json/${topic}/tasks`;
```

RutZap (approved):

